PR #41, opened 2022-09-30
Update timestamping authority to support KMS signer

haydentherapper:
Changes:
* Temporarily switch to personal fork of pkcs7/timestamping libraries. This will be updated once a change to upstream is accepted.
* api.go - Update to support KMS signer. Additionally the certificate chain is now expected to include the leaf, to avoid per-instance signing certificates.
* memory.go - Update generation of in-memory root and leaf
* x509.go - Update verifier to check additional properties
* cli - Fix output to not reference Rekor
* fetch_tsa_certs - Construct leaf certificate additionally
* server - Change flag names to use a dash


haydentherapper:
cc @asraa

haydentherapper:
cc @malancas

pkg/x509/testutils/cert_test_utils.go
```diff
@@ -115,17 +116,29 @@ func GenerateSubordinateCa(rootTemplate *x509.Certificate, rootPriv crypto.Signe
 }
 
 func GenerateLeafCert(parentTemplate *x509.Certificate, parentPriv crypto.Signer) (*x509.Certificate, *ecdsa.PrivateKey, error) {
+	timestampExt, err := asn1.Marshal([]asn1.ObjectIdentifier{{1, 3, 6, 1, 5, 5, 7, 3, 8}})
```
haydentherapper, line 119: Can't use the consts I created because it would create an import cycle

.gitignore
```diff
@@ -3,3 +3,6 @@ dist/*
 timestampCLIImagerefs
 timestampServerImagerefs
 fetchTSAImagerefs
+timestamp-cli
+timestamp-server
+*.tsr
```
cpanato, line 8: nit: new line

cmd/fetch-tsa-certs/fetch_tsa_certs.go
```diff
@@ -145,6 +148,49 @@ func fetchCACertificate(ctx context.Context, parent, kmsKey string, client *priv
...
 		parsedCerts = append(parsedCerts, certs[0])
 	}
+	intermediate := parsedCerts[0]
+
+	// generate leaf certificate
+	leafKMSSigner, err := kms.Get(ctx, leafKMSKey, crypto.SHA256)
+	if err != nil {
+		return nil, err
+	}
+	leafSigner, _, err := leafKMSSigner.CryptoSigner(ctx, func(err error) {})
```
asraa, line 158: do you need this? why no pass in `leafKMSSigner.PublicKey()` into the `x509.CreateCertificate` below?

pkg/x509/x509.go
```diff
@@ -0,0 +1,115 @@
...
+	"crypto"
+	"crypto/x509"
+	"encoding/asn1"
+	"errors"
+
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+)
+
+var (
+	EKUOID             = asn1.ObjectIdentifier{2, 5, 29, 37}
```
asraa, line 27: nit: add a comment here: this is extended key usage OID?

pkg/x509/x509.go
```diff
@@ -0,0 +1,115 @@
...
+		Intermediates: intermediates,
+		KeyUsages: []x509.ExtKeyUsage{
+			x509.ExtKeyUsageTimeStamping,
+		},
+	}
+	if _, err := certs[0].Verify(opts); err != nil {
+		return err
+	}
+
+	// Verify that all certificates but the leaf are CA certificates
```
asraa, line 62: does x509.Certificate.Verify also verify that the root was self-signed?

asraa (commented):
Seems like the change now enforces that without in-memory you must have an intermediate?

pkg/x509/x509.go
haydentherapper: updated

pkg/x509/x509.go
haydentherapper: No, and I actually reported this as a bug to golang, issue 51953. The response was that it's up to the caller to decide what is trusted as a root, but they did recognize it's a bug to not check that the root is a CA (as in, I can't pass a non-CA leaf as a trusted CA cert).

haydentherapper:
This shouldn't enforce that in the KMS key you need to provide an intermediate. Was there a particular line where you saw this? This enforces there must be two certs, but it's leaf and root at minimum.

cmd/fetch-tsa-certs/fetch_tsa_certs.go
haydentherapper: good call, updated!

asraa (approved):

Issue #38, opened 2022-09-29
Add Recoverer middleware


haydentherapper:
**Description**

https://github.com/go-chi/chi/blob/master/middleware/recoverer.go

Also review other middleware that's currently enabled

malancas:
I'll take a look at this one as well.

malancas:
@haydentherapper it looks like you already added the recoverer middleware a little while ago https://github.com/sigstore/timestamp-authority/blob/main/pkg/generated/restapi/configure_timestamp_server.go#L112

I can still review the already existing middleware in this issue though.

PR #59, opened 2022-10-07
remove unused middleware

malancas:
#### Summary

Closes https://github.com/sigstore/timestamp-authority/issues/38.

The recoverer middleware was already implemented in the code but I reviewed the remaining middleware. I removed unused middleware in this PR. The middleware removed was supporting an endpoint that is not documented by the openapi spec.

### DIFF
--- a/pkg/generated/restapi/configure_timestamp_server.go
+++ b/pkg/generated/restapi/configure_timestamp_server.go
@@ -71,7 +71,7 @@ func configureAPI(api *operations.TimestampServerAPI) http.Handler {
 
 	api.ServerShutdown = func() {}
 
-	api.AddMiddlewareFor("GET", "/api/v1/timestamp", middleware.NoCache)
+	api.AddMiddlewareFor("POST", "/api/v1/timestamp", middleware.NoCache)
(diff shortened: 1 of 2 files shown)


pkg/generated/restapi/configure_timestamp_server.go
```diff
@@ -71,7 +71,6 @@ func configureAPI(api *operations.TimestampServerAPI) http.Handler {
 
 	api.ServerShutdown = func() {}
 
-	api.AddMiddlewareFor("GET", "/api/v1/timestamp", middleware.NoCache)
 	api.AddMiddlewareFor("GET", "/api/v1/log/timestamp/certchain", cacheForDay)
```
haydentherapper, line 74: I think we need to keep `/api/v1/timestamp`, but update the second one, since the new API path is `/api/v1/timestamp/certchain`

malancas: Should I update the first one to handle POST requests? I don't see a GET `/api/v1/timestamp` endpoint in the openapi spec.

haydentherapper: Yep! Since we don’t persist timestamps, it’s only POST

haydentherapper (approved):

Issue #3, opened 2022-07-13
Choose timestamp signer


haydentherapper:
In-memory backed by KMS (by Tink) is my preference.

Other option would be KMS.

haydentherapper:
* Tink for high performance (in memory) but secured key
* KMS for alternative for key storage
* In memory for testing
* Password protected key for testing in Scaffolding

haydentherapper:
KMS and in-memory are done.

We should add a file based key for Scaffolding test.

PR #64, opened 2022-10-11
Add Tink and File-based signers

haydentherapper:
#### Summary

This adds signing backends for Tink (to enable in-memory signing while securely protecting the signing key with KMS) and file-backed signers (for Scaffolding).

This copies in all of the Tink code from Fulcio. Ideally I'll move this into sigstore at some point so we can reuse it (and in Rekor!). Also copied the file signer code from Rekor.

Fixes https://github.com/sigstore/timestamp-authority/issues/3


haydentherapper:
cc @vaikas fyi for the file-backed signer to support scaffolding

cc @malancas

cmd/timestamp-server/app/root.go
```diff
@@ -64,10 +64,11 @@ func init() {
...
 	rootCmd.PersistentFlags().String("timestamp-signer", "memory", "Timestamping authority signer. Valid options include: [kms, tink, memory, file]. Memory and file-based signers should only be used for testing")
 	// KMS flags
-	rootCmd.PersistentFlags().String("kms-key-resource", "memory", "KMS key for signing timestamp responses. Valid options include: [gcpkms://resource, azurekms://resource, hashivault://resource, awskms://resource]")
+	rootCmd.PersistentFlags().String("kms-key-resource", "", "KMS key for signing timestamp responses. Valid options include: [gcpkms://resource, azurekms://resource, hashivault://resource, awskms://resource]")
 	// Tink flags
-	rootCmd.PersistentFlags().String("tink-key-resource", "memory", "KMS key for signing timestamp responses for Tink keysets. Valid options include: [gcp-kms://resource, aws-kms://resource]")
-	rootCmd.PersistentFlags().String("tink-keyset-path", "memory", "Path to KMS-encrypted keyset for Tink, decrypted by tink-key-resource")
+	rootCmd.PersistentFlags().String("tink-key-resource", "", "KMS key for signing timestamp responses for Tink keysets. Valid options include: [gcp-kms://resource, aws-kms://resource]")
```
malancas, line 69: In the `kms-key-resource` flag options, `gcpkms://resource` and `awskms://resource` don't have a dash, like `gcp-kms://resource` in this flag option. Can we standardize on both flags either using dashes or not using dashes in the name?

haydentherapper: That's a fair point. The former came from sigstore/sigstore (and I think was inspired by the industry), the latter is how Tink chose to specify the keys.

One concern is you'd use something like tinkey (https://github.com/google/tink/blob/27b061bb9ed1af1a6f538410bff443290e427e66/docs/TINKEY.md) to generate the keyset, you'd then be specifying a different key prefix for the server, which might be confusing. Thoughts on this? It's easy enough to handle the different prefix in the server if we want to unify the two, but might confuse a regular Tink user.

pkg/signer/file.go
```diff
@@ -0,0 +1,61 @@
...
+			return nil, err
+		}
+		return &File{signer}, nil
+	case *ecdsa.PrivateKey:
+		signer, err := signature.LoadECDSASignerVerifier(pk, crypto.SHA256)
+		if err != nil {
+			return nil, err
+		}
+		return &File{signer}, nil
+	case ed25519.PrivateKey:
```
vaikas, line 52: Just curious why this is not a pointer when the others are? If it's intentional, mayhaps a comment? If it's common knowledge and only I'm confused, then fine :)

vaikas: Maybe add a test case for this as well in file_test?

vaikas (commented):
Sweet!

cmd/timestamp-server/app/root.go
malancas: I'm leaning towards using the Tink prefix format for both flags because of the tinkey keyset generation situation you mentioned and because it looks like the other flag format isn't used anywhere else in the code really.

haydentherapper: The rest of sigstore uses the non-hyphenated one, so we should probably prefer that over Tink. I'm somewhat leaning towards keeping them as-is, because there's other discrepancies besides hyphenation (hcvault vs hashicorp)

A longterm solution could be proposing or updating an existing [AIP](https://google.aip.dev/) to specify how to reference a key by provider, to align sigstore, Tink, and other libraries.

malancas: Leaving this as is and readdressing in the longterm with an AIP makes sense to me.

pkg/signer/file.go
haydentherapper: `ParsePKCS8PrivateKey` "returns a *rsa.PrivateKey, a *ecdsa.PrivateKey, or a ed25519.PrivateKey". I'm not sure why the decision was made to make only ed25519 not a pointer, but it's permeated through golang crypto.

Added a test, and switched to generating keys in-memory!

haydentherapper:
@asraa @malancas Bump for LGTM

malancas (approved):

Issue #5, opened 2022-07-13
Add monitoring


haydentherapper:

malancas:
Can't assign myself but I'll take this

haydentherapper:
Weird, your access got removed entirely when switching from triage to write. Added you back, you should be good to go!

malancas:
Great, thank you!

PR #67, opened 2022-10-12
Add monitoring

malancas:
#### Summary

Closes https://github.com/sigstore/timestamp-authority/issues/5.

This PR creates some starting prometheus metrics and add them to the `wrapMetrics` middleware. I checked that the metrics are reporting as expected on the prometheus endpoint.

We can create follow up issues around how we want to capture these metrics.

### DIFF
--- a/pkg/api/metrics.go
+++ b/pkg/api/metrics.go
@@ -24,7 +24,7 @@ import (
 
 var (
 	MetricLatency = promauto.NewHistogramVec(prometheus.HistogramOpts{
-		Name: "timestamp_api_latency",
+		Name: "timestamp_authority_api_latency",
 		Help: "API Latency on calls",
 	}, []string{"path", "code"})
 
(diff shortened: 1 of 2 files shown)


pkg/api/metrics.go
```diff
@@ -16,13 +16,34 @@
...
 import (
+	"time"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 var (
 	MetricLatency = promauto.NewHistogramVec(prometheus.HistogramOpts{
 		Name: "timestamp_api_latency",
```
haydentherapper, line 27: Can we standardize the prefix on the metrics for all metrics? "tsa" or "ts" or "timestamp" or "timestamp_authority"?

haydentherapper (commented):
Looks great!

pkg/api/metrics.go
malancas: Just prefixed all the metrics with "timestamp_authority".

haydentherapper (approved):

Issue #73, opened 2022-10-13
Verification fails with libressl


hectorj2f:
**Description**

I am investigating why the verfication is failing when following the README steps. I tried both local deployment approaches and both are failing.

```
Verification: FAILED
4336174636:error:2FFFF065:time stamp routines:CRYPTO_internal:ess signing certificate error:/AppleInternal/Library/BuildRoots/a0876c02-1788-11ed-b9c4-96898e02b808/Library/Caches/com.apple.xbs/Sources/libressl/libressl-2.8/crypto/ts/ts_rsp_verify.c:300:
```

where the timestamp inspection shows:

```
Status info:
Status: Granted.
Status description: unspecified
Failure info: unspecified

TST info:
Version: 1
Policy OID: 0.4.0.2023.1.1
Hash Algorithm: sha256
Message data:
    0000 - e4 9a 89 7a ac 4c 98 19-21 27 0a 3f 79 e4 68 85   ...z.L..!'.?y.h.
    0010 - c9 95 2f c5 ba 1a b4 c0-cd ef 38 0b e7 73 74 81   ../.......8..st.
Serial number: 0x1F5E0336A9118E4020B3EC84C33B2C9E99E36505
Time stamp: Oct 13 17:53:10 2022 GMT
Accuracy: 0x01 seconds, unspecified millis, unspecified micros
Ordering: no
Nonce: 0x4CE5002F5E5F753A58F2B98124C3B6A732EE769B
TSA: DirName:/O=local/CN=Test TSA Timestamping
Extensions:
```

haydentherapper:
Hm, I wonder if this has to do with either macOS overriding the trusted root or libressl vs openssl.

haydentherapper:
To confirm, to verify you ran `openssl ts -verify -in response.tsr -data "myblob" -CAfile ts_chain.pem`?

haydentherapper:
In terms of verification, what should happen is as follows:
* The leaf certificate is extracted from the response
* The chain must be provided
* openssl uses `CAfile` for the root
* For the example, the root should verify the leaf
* Success!

I think we may need an extra flag if you're using an intermediate, but for the example, it should just be a leaf and root.

hectorj2f:
Yes, I ran `openssl ts -verify -in response.tsr -data "myblob" -CAfile ts_chain.pem`

haydentherapper:
[...]

PR #75, opened 2022-10-14
chore: add note about openssl on macos

hectorj2f:

#### Summary

closes https://github.com/sigstore/timestamp-authority/issues/73

#### Release Note

Add note in README to run example on MacOS


haydentherapper (approved):
Thanks for looking into this! I'm really curious what's different, there's probably a bug in libressl then.

hectorj2f:
I invested a reduced amount of time investigating but I found several issues related to the verify command here https://github.com/libressl-portable/portable.

Issue #76, opened 2022-10-14
[timestamp-server] Serving over HTTPS is not configurable


hectorj2f:
**Description**

It seems the server code supports the usage of https, but it is currently hardcoded to use `http`, https://github.com/sigstore/timestamp-authority/blob/main/cmd/timestamp-server/app/serve.go#L66. In addition, I found many of the cobra flags weren't bound to the viper flags. To figure it out, I passed flags `--scheme` to the server but that one was not bound to the viper flags.

I'm also wondering whether it makes sense to have HTTPS support for the TSA.

PR #77, opened 2022-10-14
Allow serving over http or https

hectorj2f:

#### Summary

While looking at the code, I found the server was only serving over http. I'm not sure if this is purely intentional although I could think of private TSAs.

closes: https://github.com/sigstore/timestamp-authority/issues/76

### DIFF
--- a/cmd/timestamp-server/app/serve.go
+++ b/cmd/timestamp-server/app/serve.go
@@ -38,7 +38,9 @@ var serveCmd = &cobra.Command{
 	Short: "start http server with configured api",
 	Long:  `Starts a http server and serves the configured api`,
 	Run: func(cmd *cobra.Command, args []string) {
-
+		if err := viper.BindPFlags(cmd.Flags()); err != nil {
+			log.Logger.Fatal(err)
+		}
 		// Setup the logger to dev/prod
 		log.ConfigureLogger(viper.GetString("log-type"))
 
@@ -63,7 +65,7 @@ var serveCmd = &cobra.Command{
 
 		server.Host = viper.GetString("address")
 		server.Port = int(viper.GetUint("port"))
-		server.EnabledListeners = []string{"http"}
+		server.EnabledListeners = viper.GetStringSlice("scheme")
 
 		api.ConfigureAPI()
 		server.ConfigureAPI()


cmd/timestamp-server/app/serve.go
```diff
@@ -38,7 +38,9 @@ var serveCmd = &cobra.Command{
 	Short: "start http server with configured api",
 	Long:  `Starts a http server and serves the configured api`,
 	Run: func(cmd *cobra.Command, args []string) {
-
+		if err := viper.BindPFlags(cmd.Flags()); err != nil {
```
haydentherapper, line 41: Flags are bound here - https://github.com/sigstore/timestamp-authority/blob/main/cmd/timestamp-server/app/root.go#L71

cmd/timestamp-server/app/serve.go
```diff
@@ -63,7 +65,7 @@ var serveCmd = &cobra.Command{
 
 		server.Host = viper.GetString("address")
 		server.Port = int(viper.GetUint("port"))
-		server.EnabledListeners = []string{"http"}
+		server.EnabledListeners = viper.GetStringSlice("scheme")
```
haydentherapper, line 68: FWIW, this was copied from Rekor, which only supports http.

I believe we'd need additional configuration to support HTTPS. I'd probably say to just stick to supporting HTTP and terminating the TLS connection at ingress.

cmd/timestamp-server/app/serve.go
hectorj2f: Yes, for these flags, if you try using any other available flags those are not bound, e.g. --scheme

cmd/timestamp-server/app/serve.go
hectorj2f: The code generator provides all the code to support https, I suggest you run `timestamp-server serve -h`. Otherwise we should also exclude those available but unsupported flags.

Yes, I saw it in Rekor too. I had the feeling https makes more sense for the TSA.

cmd/timestamp-server/app/serve.go
haydentherapper: Ah thanks, thought that was for all flags. I’ll take another look Monday!

hectorj2f: So we should either remove these flags or allow users to set them when desired.

cmd/timestamp-server/app/serve.go
bobcallaway: typically you'd run with a separate ingress or proxy that terminates SSL, so thats why the listener was originally pinned to http only.

haydentherapper: Agreed, I think it's reasonable to support it optionally though. `scheme` defaults to `http` so HTTPS would only be with explicit configuration

haydentherapper (approved):

cmd/timestamp-server/app/serve.go
hectorj2f: @bobcallaway @haydentherapper Yes, that setup might be common under most of organizations. It was more of a question about wether we allow them to use these flags or delete them from the available ones.

haydentherapper:
Filed https://github.com/sigstore/timestamp-authority/issues/78 to clean up duplicated flags

hectorj2f:
@haydentherapper Awesome, thanks :).

Yes, I saw the issue 👏🏻

haydentherapper:
Thanks for pointing this out @hectorj2f! I wasn't even aware of these auto-generated flags. One of the downsides of openapi I guess.

Issue #78, opened 2022-10-17
Clean up duplicated flags


haydentherapper:
**Description**

* `hostname` is unused
* `address` is the same as the openapi provided `host` flag
* `port` is duplicated
* read and write timeouts should be set by `read-timeout`/`write-timeout` for pprof/metrics

PR #84, opened 2022-10-18
Clean up duplicate flags

malancas:
#### Summary

Closes https://github.com/sigstore/timestamp-authority/issues/78.

### DIFF
--- a/cmd/timestamp-cli/app/timestamp.go
+++ b/cmd/timestamp-cli/app/timestamp.go
@@ -37,7 +37,7 @@ import (
 	"github.com/spf13/viper"
 )
 
-func addTimestampFlags(cmd *cobra.Command) error {
+func addTimestampFlags(cmd *cobra.Command) {
 	cmd.Flags().Var(NewFlagValue(fileFlag, ""), "artifact", "path to an artifact to timestamp")
 	cmd.Flags().String("hash", "sha256", "hash algorithm to use - Valid values are sha256, sha384, and sha512")
 	cmd.Flags().Bool("nonce", true, "specify a pseudo-random nonce in the request")
@@ -45,7 +45,6 @@ func addTimestampFlags(cmd *cobra.Command) error {
 	cmd.Flags().Var(NewFlagValue(oidFlag, ""), "tsa-policy", "optional dotted OID notation for the policy that the TSA should use to create the response")
 
 	cmd.Flags().String("out", "response.tsr", "path to a file to write response.")
-	return nil
 }
 
 func validateTimestampFlags() error {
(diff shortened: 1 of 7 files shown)


cmd/timestamp-cli/app/timestamp.go
```diff
@@ -37,15 +37,14 @@ import (
 	"github.com/spf13/viper"
 )
 
-func addTimestampFlags(cmd *cobra.Command) error {
```
malancas, line 40: This function never calls any functions that could produce errors or creates an error itself, so I removed this return type.

cmd/timestamp-server/app/root.go
```diff
@@ -60,8 +62,10 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&logType, "log-type", "dev", "logger type to use (dev/prod)")
 	rootCmd.PersistentFlags().BoolVar(&enablePprof, "enable-pprof", false, "enable pprof for profiling on port 6060")
 
-	rootCmd.PersistentFlags().String("hostname", "timestamp.sigstore.dev", "public hostname of instance")
-	rootCmd.PersistentFlags().String("address", "127.0.0.1", "Address to bind to")
+	// pprof/metrics read and write timeout flags
+	rootCmd.PersistentFlags().DurationVar(&readTimeout, "read-timeout", 10*time.Second, "pprof/metrics read timeout")
```
haydentherapper, line 66: there's an existing read-timeout/write-timeout flag provided - https://github.com/sigstore/timestamp-authority/blob/main/pkg/generated/restapi/server.go#L100-L101

Can we reuse these?

malancas: Yes, good catch.

cmd/timestamp-server/app/root.go
```diff
@@ -75,34 +73,9 @@ func init() {
...
 
-	rootCmd.PersistentFlags().Uint16("port", 3000, "Port to bind to")
-
 	if err := viper.BindPFlags(rootCmd.PersistentFlags()); err != nil {
 		log.Logger.Fatal(err)
 	}
-
-	log.Logger.Debugf("pprof enabled: %v", enablePprof)
-	// Enable pprof
-	if enablePprof {
```
malancas, line 86: After some locally running the server with `--enable-pprof=true`, I don't think this block will ever be entered. The default `false` value of `enablePprof` is always used because the value of the flag is never evaluated via viper with `viper.GetBool("enable-pprof")`.

To solve this, I created a new `server` package that contains a pprof server constructor. This constructor is called in `timestamp-server/app/serve.go` after the flag's value is evaluated in that file.

I would like to create a constructor for the restapi server in this new package and call that in `timestamp-server/app/serve.go` as well but that can be done in a follow up PR if preferred to keep this one smaller.

haydentherapper: Good catch! interested as to why the value is never evaluated.

haydentherapper (approved):

malancas:
@haydentherapper thanks for reviewing, would you mind merging this?

haydentherapper:
@malancas Sure, I wanted to leave it open for a little bit in case @hectorj2f had any other comments.

malancas:
Gotcha, that sounds good.

haydentherapper (approved):

PR #92, opened 2022-10-20
Update timestamp dependency

malancas:
#### Summary

Noticed that the `POST /api/v1/timestamp` endpoint was failing due to a change in the timestamp dependency. This updates the dependency which fixes the issue.


haydentherapper (approved):

haydentherapper:
Thanks! I guess this grabbed the previous commit hash accidentally.

Issue #39, opened 2022-09-30
Improve timestamp-cli


haydentherapper:
**Description**

* Review the output
* Review the input params
* Make sure there's no leftovers from Rekor
* Add timestamp verification into CLI

haydentherapper:
Changing to a p0 - I'm going to define p0 as something we want to get done before an initial release of the project.

haydentherapper:
Remaining task is to add an inspect flag

PR #98, opened 2022-10-24
Add inspect cli cmd

malancas:
#### Summary

Closes https://github.com/sigstore/timestamp-authority/issues/39 by adding an `inspect` command to the CLI that enables the user to inspect a timestamp response.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 
 .PHONY: all test clean clean-gen lint gosec ko ko-local
 
-all: timestamp-cli timestamp-server
+all: bin/timestamp-cli bin/timestamp-server
 
 GENSRC = pkg/generated/client/%.go pkg/generated/models/%.go pkg/generated/restapi/%.go
 OPENAPIDEPS = openapi.yaml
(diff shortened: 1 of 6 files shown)


malancas:
@haydentherapper do you have any reservations about building `timestamp-server` and `timestamp-cli` in a `bin` directory instead of in the root directory? I ask because the current `timestamp-cli` entry in `.gitignore` is causing Git to ignore any changes in the `cmd/timestamp-cli` directory.

Makefile
```diff
@@ -69,19 +69,19 @@ gosec:
 
 gen: $(GENSRC)
 
-timestamp-cli: $(SRCS)
-	CGO_ENABLED=0 go build -trimpath -ldflags "$(CLI_LDFLAGS)" -o timestamp-cli ./cmd/timestamp-cli
+bin/timestamp-cli: $(SRCS)
```
haydentherapper, line 72: Can we leave these with the bin prefix? I think so, this is just how to refer to it in other targets if i recall makefiles

cmd/timestamp-cli/app/inspect.go
```diff
@@ -0,0 +1,77 @@
...
+	cmd.Flags().Var(NewFlagValue(formatFlag, ""), "format", "output format")
+	cmd.MarkFlagRequired("format") //nolint:errcheck
+}
+
+type inspectCmdOutput struct {
+	TimestampResponse timestamp.Timestamp
+}
+
+func (t *inspectCmdOutput) String() string {
+	return fmt.Sprintf("%+v", t.TimestampResponse)
```
haydentherapper, line 42: LGTM - We might want to evolve this over time to be structured differently, but I think this is good for now!

haydentherapper (approved):

Issue #2, opened 2022-07-13
Verify conformance with specifications


haydentherapper:
- [ ] https://datatracker.ietf.org/doc/html/rfc3161 - Main RFC for timestamping
- [ ] https://datatracker.ietf.org/doc/html/rfc3628 - RFC for policy requirements that we need to adhere to (we set an OID for this RFC to denote that we're conformant)

An alternative policy for Apple's TSA: https://images.apple.com/certificateauthority/pdf/Apple_Timestamp_CPS_v2.0.pdf

We may need to create our own TSA policy with its own OID if we can't conform to 3628

malancas:
Github won't let me assign myself to the issue yet but I can start taking a look at this.

haydentherapper:
https://datatracker.ietf.org/doc/html/rfc3628#section-7.2 and https://datatracker.ietf.org/doc/html/rfc3628#section-7.4 are the main sections on policy that I'm not sure we can conform to in a cloud environment. MS or Apple might have a better TSA policy that's more up to date with modern deployment.

malancas:
After reading through [https://datatracker.ietf.org/doc/html/rfc3628](https://datatracker.ietf.org/doc/html/rfc3628) and specifically sections [7.2](https://datatracker.ietf.org/doc/html/rfc3628#section-7.2) and [7.4](https://datatracker.ietf.org/doc/html/rfc3628#section-7.4) noted above, I don't believe we can conform to in a cloud environment. These sections discuss the generation of keys in a physically secure location and securing/limiting the access to the facilities that host the keys and TSA hardware, see [section 7.4.4](https://datatracker.ietf.org/doc/html/rfc3628#section-7.4.4).

Apple has a [timestamp sub-CA certification practice statement](https://images.apple.com/certificateauthority/pdf/Apple_Timestamp_CPS_v2.0.pdf). Section 3.2.1 covers private key storage and specifies that the private key must be stored in a HSM validated at a minimum level of FIPS 140-2 Level 3. GCP offers a [cloud-based HSM service](https://cloud.google.com/kms/docs/hsm) that fulfills the same FIPS requirement.

However the statement also discusses physical security in section 5.5. This section includes defining "security perimeters with appropriate physical barriers to entry around the business premises and Timestamp Sub-CA facilities". Because we are using a cloud environment, this is not something we can enforce ourselves.

malancas:
I'll look into whether MS has a more modern one as well.

[...]

PR #100, opened 2022-10-24
Update oid

malancas:

#### Summary

Update the OID to match the one specified in https://github.com/sigstore/timestamp-authority/issues/2.

### DIFF
--- a/pkg/api/timestamp.go
+++ b/pkg/api/timestamp.go
@@ -39,8 +39,7 @@ func TimestampResponseHandler(params ts.GetTimestampResponseParams) middleware.R
 
 	policyID := req.TSAPolicyOID
 	if policyID.String() == "" {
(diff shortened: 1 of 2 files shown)


pkg/api/timestamp.go
```diff
@@ -39,8 +39,7 @@ func TimestampResponseHandler(params ts.GetTimestampResponseParams) middleware.R
 
 	policyID := req.TSAPolicyOID
 	if policyID.String() == "" {
-		// https://datatracker.ietf.org/doc/html/rfc3628#section-5.2
-		policyID = asn1.ObjectIdentifier{0, 4, 0, 2023, 1, 1}
+		policyID = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 57264, 2}
```
haydentherapper, line 42: There should be a test that breaks with this change too

haydentherapper (approved):

PR #118, opened 2022-11-04
Fix hashed message verification

hectorj2f:
#### Summary

When trying to time-stamp messages longer than 32 bytes, I realized the verification function was failing to verify the response and the input message hashes. I was seen the following error: `Hashed messages don't match`.

Then I had a look at the `digitorus/timestamp` implementation and I observed that messages longer than 32 bytes were timestamped without issues. So I had a look at our verification code and found a different way of validating the hashed messages than the one used in the timestamp code.

#### Release Note

fix: verify hashed messages


cmd/timestamp-cli/app/verify.go
```diff
@@ -130,22 +130,18 @@ func verifyArtifactWithTSR(ts *timestamp.Timestamp) error {
...
-	}
+	return verifyHashedMessages(ts.HashAlgorithm.New(), ts.HashedMessage, artifactBytes)
+}
 
-	_, err = h.Write(b[:n])
+func verifyHashedMessages(hashAlg hash.Hash, hashedMessage []byte, artifactBytes []byte) error {
+	h := hashAlg
+	_, err := h.Write(artifactBytes)
 	if err != nil {
 		return fmt.Errorf("Failed to create local message hash")
```
bobcallaway, line 140:
```suggestion
		return fmt.Errorf("Failed to create local message hash: %w", err)
```

cmd/timestamp-cli/app/verify_test.go
```diff
@@ -0,0 +1,86 @@
...
+	for _, tc := range tests {
+		tsq, err := timestamp.CreateRequest(strings.NewReader(tc.message), &timestamp.RequestOptions{
+			Hash:         crypto.SHA256,
+			Certificates: true,
+		})
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		tsr, err := http.Post("https://freetsa.org/tsr", "application/timestamp-query", bytes.NewReader(tsq))
```
bobcallaway, line 62: Do you really need to call out to an external TSA? If you could do this without the network call, that would be ideal.

cmd/timestamp-cli/app/verify.go
```diff
@@ -130,22 +130,18 @@ func verifyArtifactWithTSR(ts *timestamp.Timestamp) error {
...
-	r := bytes.NewReader(artifactBytes)
-	n, err := r.Read(b)
-	if err == io.EOF {
-		return err
-	}
+	return verifyHashedMessages(ts.HashAlgorithm.New(), ts.HashedMessage, artifactBytes)
+}
 
-	_, err = h.Write(b[:n])
+func verifyHashedMessages(hashAlg hash.Hash, hashedMessage []byte, artifactBytes []byte) error {
```
bobcallaway, line 136:
```suggestion
func verifyHashedMessages(hashAlg hash.Hash, hashedMessage []byte, artifactReader io.Reader) error {
```

While TSRs aren't likely very big, there's no need to copy around []byte slices if you're just taking a pass over it to compute a hash and not doing anything else

hectorj2f: Sure, done.

cmd/timestamp-cli/app/verify_test.go
hectorj2f: Changed the code to reuse the local server.

cmd/timestamp-cli/app/verify.go
```diff
@@ -130,22 +131,27 @@ func verifyArtifactWithTSR(ts *timestamp.Timestamp) error {
...
+	for {
+		n, err := artifactReader.Read(b)
+		if err == io.EOF {
+			break
+		}
+
+		_, err = h.Write(b[:n])
+		if err != nil {
+			return fmt.Errorf("failed to create hash %w", err)
+		}
```
bobcallaway, line 150:
```suggestion
	if _, err := io.Copy(h, artifactReader); err != nil {
		return fmt.Errorf("failed to create hash %w", err)
	}
```

bobcallaway: h is an `io.Writer` so you can just let the `io` library do the buffering internally.

cmd/timestamp-cli/app/verify.go
```diff
@@ -130,22 +131,17 @@ func verifyArtifactWithTSR(ts *timestamp.Timestamp) error {
...
 
-	h := ts.HashAlgorithm.New()
-	b := make([]byte, h.Size())
-
-	r := bytes.NewReader(artifactBytes)
-	n, err := r.Read(b)
-	if err == io.EOF {
-		return err
-	}
+	return verifyHashedMessages(ts.HashAlgorithm.New(), ts.HashedMessage, bytes.NewReader(artifactBytes))
```
bobcallaway, line 134: if you change line 128 to just `os.Open` then you can pass the `*os.File` as it implements io.Reader

hectorj2f:
@bobcallaway Thanks for the review

cpanato (approved):

haydentherapper (commented):
Good catch, thanks for fixing! Looks like the root cause was reading into a buffer of size(sha256 hash) bytes

PR #140, opened 2022-11-09
fix setting certChain.

vaikas:
Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

#### Summary
Fix a bug where certChain was being recreated, therefore setting it in the if clause being invisible. Basically it doesn't work on that path.

#### Release Note
* Fix a bug where certChain was not set correctly.


haydentherapper (approved):
Whoops! I thought I had confirmed this working with Tink!

Issue #133, opened 2022-11-08
Integrating with a NTP server


malancas:
**Description**

We should investigate the possibility of integrating the TSA with a NTP server. [CloudFlare offers a free NTP server](https://blog.cloudflare.com/secure-time/) as a possible option (credit to @haydentherapper for finding this). GCP also offers [NTP integration for its VMs](https://cloud.google.com/compute/docs/instances/configure-ntp).

In addition to providing a NTP server, the CloudFlare article also discusses an updated protocol based on NTP called NTS to communicate with the server. However, there does not seem to be Go support for the NTS protocol. There seem to be a few Go implementations of the NTP protocol on GitHub.

kommendorkapten:
Nice, I was not aware of this issue.

I suggest we keep this PR https://github.com/sigstore/timestamp-authority/pull/143 as is, then continue to research state of NTS separately. Some more details on NTS: https://docs.ntpsec.org/latest/NTS-QuickStart.html

kommendorkapten:
And also this: https://www.netnod.se/nts/network-time-security

haydentherapper:
Agreed, let’s scope this issue to just NTP integration for your PR and we can follow up with another issue for NTS.

haydentherapper:
Closing now that https://github.com/sigstore/timestamp-authority/pull/143 is done - We chatted about NTS and didn't see a benefit over NTP for monitoring.

PR #143, opened 2022-11-11
Added an optional feature to compare the local time with a set of trusted ntp servers

kommendorkapten:
If the local time is diffing more than a configured threshold, the service shuts down itself.

Signed-off-by: Fredrik Skogman <kommendorkapten@github.com>

#### Release Note
* Optional comparison of service's local time with a set of ntp servers, to make sure local time is not drifting.


kommendorkapten:
The motivation behind this PR is that for a TSA it's _extremely_ important that we are using a time source that's accurate. Just relying on the host's time can be dangerous, as NTP may not be enabled, or the configured NTP server is not correct. Having the process itself verifying its time with external thirdparty servers is defence in depth, and following the "trust but verify" philosophy.

cmd/timestamp-server/app/serve.go
```diff
@@ -80,6 +81,31 @@ var serveCmd = &cobra.Command{
...
+				for {
+					err = ntpm.Start()
+					if err != nil {
+						log.Logger.Warnf("ntp monitor: %s", err)
+					}
+					// Local time have drifted from the
+					// observed NTP time. Shut down the
+					// service.
+					if err == ntpmonitor.ErrInvTime {
+						log.Logger.Fatalf("local time can not be trusted: %s", err)
```
haydentherapper, line 103: Can we record a metric rather than crash? I'd worry some brief outage or misconfiguration in an NTP server would result in a sigstore outage.

ntpsync.yaml
```diff
@@ -0,0 +1,87 @@
...
+  - "time.apple.com"
+
+  # Google AS15169
+  - "time.google.com"
+  - "time1.google.com"
+  - "time2.google.com"
+  - "time3.google.com"
+  - "time4.google.com"
+
+  # Facebook AS32934
```
haydentherapper, line 42: I'm gonna be nitpicky, can we remove fb from the list? You can ask offline why :)

cmd/timestamp-server/app/serve.go
```diff
@@ -80,6 +81,31 @@ var serveCmd = &cobra.Command{
...
+				ntpm, err := ntpmonitor.New(ntpMonitoring)
+				if err != nil {
+					log.Logger.Fatalf("error initializing ntp monitor %s", err)
+				}
+
+				// Restart ntp monitor if needed
+				for {
+					err = ntpm.Start()
+					if err != nil {
+						log.Logger.Warnf("ntp monitor: %s", err)
```
haydentherapper, line 97: Same here, every failure should result in a recorded metrics imo

pkg/ntpmonitor/ntpmonitor.go
```diff
@@ -0,0 +1,151 @@
...
+	return &NTPMonitor{cfg: cfg}
+}
+
+// Start the periodic monitor. If there is an initialization error, the
+// function returns immediately. Once the periodic monitoring starts, it runs
+// until an error is found, or Stop() is being called.
+func (n *NTPMonitor) Start() error {
+	n.run.Store(true)
+
+	if len(n.cfg.Servers) < n.cfg.NumServers {
```
haydentherapper, line 68: Can configuration errors be checked on initialization (New()) rather than in Start()?

pkg/ntpmonitor/ntpmonitor.go
```diff
@@ -0,0 +1,151 @@
...
+
+		// Local time is in sync. Wait for next poll.
+		time.Sleep(time.Duration(n.cfg.Period) * time.Second)
+	}
+
+	return nil
+}
+
+// Stop the monitoring.
+func (n *NTPMonitor) Stop() {
```
haydentherapper, line 127: Do we need to call Stop on server shutdown?

haydentherapper (commented):
One thing we discussed in https://github.com/sigstore/timestamp-authority/issues/133 was using NTP servers that provided NTS. I'm not sure if there are enough, but something to consider, it would be great if we could secure the NTP connections

pkg/ntpmonitor/ntpmonitor.go
```diff
@@ -0,0 +1,151 @@
...
+			now := time.Now()
+
+			if err != nil {
+				log.Logger.Warnf("ntp response timeout from %s",
+					srv)
+				noResponse++
+				continue
+			}
+
+			if resp.Time.After(now.Add(-delta)) &&
```
haydentherapper, line 102: As we chatted about offline, we should confirm that there are no concerns calling NTP servers that are in another region, that NTP accounts for network latency

pkg/ntpmonitor/randomchoice.go
```diff
@@ -0,0 +1,55 @@
...
+package ntpmonitor
+
+import (
+	"math/rand"
+)
+
+// RandomChoice returns a random selection of n items from the slic s.
+// The choice is made using a PSEUDO RANDOM selection.
+// If n is greater than len(s), an empty slice is returned.
+func RandomChoice[T any](s []T, n int) []T {
```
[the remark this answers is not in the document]
kommendorkapten: Haha, yeah it's the first generics code I wrote, and actually felt natural to make this function like this :)

pkg/ntpmonitor/ntpmonitor.go
kommendorkapten: Updated.

pkg/ntpmonitor/ntpmonitor.go
kommendorkapten: Verified (as long as the rx and tx latency are ~similar). Added a comment about it.

pkg/ntpmonitor/ntpmonitor.go
kommendorkapten: Fixed.

pkg/ntpmonitor/randomchoice.go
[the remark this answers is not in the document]
haydentherapper: I think this is the first generics code I've seen in sigstore repos!

cmd/timestamp-server/app/serve.go
```diff
@@ -80,11 +81,29 @@ var serveCmd = &cobra.Command{
...
 
+		ntpMonitoring := viper.GetString("ntp-monitoring")
+		var ntpm *ntpmonitor.NTPMonitor
+		if ntpMonitoring != "" {
+			log.Logger.Infof("ntp monitoring: %s", ntpMonitoring)
+			go func() {
+				ntpm, err = ntpmonitor.New(ntpMonitoring)
+				if err != nil {
+					log.Logger.Errorf("error initializing ntp monitor %s", err)
+					return
```
haydentherapper, line 92: It'd be good to call `Fatalf` here to halt initialization since this is just on server startup

pkg/ntpmonitor/config.go
```diff
@@ -0,0 +1,52 @@
...
+			path, err)
+	}
+
+	var cfg Config
+	if err = yaml.Unmarshal(data, &cfg); err != nil {
+		return nil, fmt.Errorf("failed to parse YAML: %w", err)
+	}
+
+	return &cfg, nil
+}
```
haydentherapper, line 52: nit, can you add a test for loading from yaml too?

haydentherapper (commented):
It looks great, just a few nits!

pkg/ntpmonitor/ntpmonitor.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	return &NTPMonitor{cfg: cfg}, nil
+}
+
+// Start the periodic monitor. Once started, it runs until Stop() is called,
+func (n *NTPMonitor) Start() {
+	n.run.Store(true)
+
+	if n.cfg.RequestTimeout < 1 {
```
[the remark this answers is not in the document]
kommendorkapten: Being explicit is alway better. I'll change this.

pkg/ntpmonitor/ntpmonitor.go
```diff
@@ -0,0 +1,189 @@
...
+
+// Stop the monitoring.
+func (n *NTPMonitor) Stop() {
+	log.Logger.Info("stopping ntp monitoring")
+	n.run.Store(false)
+}
+
+// QueryNTPServer queries a provided ntp server, trying up to a configured
+// amount of times. There is one second sleep between each attempt.
+func (n *NTPMonitor) QueryNTPServer(srv string) (*ntp.Response, error) {
```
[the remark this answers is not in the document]
kommendorkapten: Done: https://github.com/sigstore/timestamp-authority/issues/148

haydentherapper (approved):

Issue #121, opened 2022-11-07
Refactor verification function with Opts struct


haydentherapper:
**Description**

Per RFC 3161 2.2, we need to support the following for verification:
- [x] Verify that the TSR's hashed message matches the digest of the artifact to be timestamped
- [x] Verify the TSR's `ESSCertID`, the certificate identifier, matches a provided TSA certificate (kind of done [here](https://github.com/digitorus/pkcs7/blob/21b8b40e6bb42e43036336c7828a1bee77c2b1d6/verify.go#L116), but see below for note on updating to support out-of-band provided leaf)
- [x] Verify the leaf certificate's subject and/or subject alternative name matches a provided subject - See https://github.com/openssl/openssl/blob/master/crypto/ts/ts_rsp_verify.c#L536
- [x] If embedded in the TSR, verify the TSR's leaf certificate matches a provided TSA certificate
- [x] Verify the TSA certificate using a CA certificate chain
- [x] Verify the TSA certificate and the intermediates (called "EKU chaining") all have the extended key usage set to only extended key usage - See https://github.com/sigstore/timestamp-authority/blob/main/pkg/x509/x509.go#L35 - Note that the pkcs7 library [verify](https://github.com/digitorus/pkcs7/blob/21b8b40e6bb42e43036336c7828a1bee77c2b1d6/verify.go#L244) does not check this
- [x] Verify the signature of the TSR using the public key in the leaf certificate (like [here](https://github.com/digitorus/pkcs7/blob/21b8b40e6bb42e43036336c7828a1bee77c2b1d6/verify.go#L111))
- [x] Verify the OID of the TSR matches an expected OID
- [x] Verify the status of the TSR does not contain an error
- [x] Verify the nonce - Mostly important for when the response is first returned

Note that the leaf certificate may be provided out of band, as it is not a requirement that the certificate is embedded in the TSR. `verifySignature` from the [pkcs7 library](https://github.com/digitorus/pkcs7/blob/21b8b40e6bb42e43036336c7828a1bee77c2b1d6/verify.go#L116) seems to expect that the certificate is present always. We'll need upstream changes I think to update verification to support TSRs without certificates.

[...]

PR #153, opened 2022-11-17
Start adding more verification with VerificationOpts struct

malancas:
#### Summary

Part of https://github.com/sigstore/timestamp-authority/issues/121. This PR starts adding additional verification using a VerificationOpts struct. More verification will be added in follow ups after some upstream changes to the timestamp package.

This PR adds the following verifications:
- Verify the TSR's certificate identifier matches a provided TSA certificate
- Verify the leaf certificate's subject name matches a provided subject
- Verify that the TSR's embedded leaf certificate matches a provided TSA certificate
- Verify the TSA certificate and the intermediates have the extended key usage set to only timestamping
- Verify the OID of the TSR matches an expected OID
- Verify the status of the TSR does not contain an error
- Verify the TSR's nonce matches a provided nonce


cmd/timestamp-cli/app/verify.go
```diff
@@ -85,6 +97,62 @@ func runVerify() (interface{}, error) {
 		return nil, err
 	}
 
+	opts, err := verification.NewVerificationOpts(ts, artifact, pemBytes)
```
malancas, line 100: @haydentherapper curerntly all verification is handled in a single function `VerifyTimestampResponse` exported from the `verification` package. Here I add new verification functions separately. Do you ahve thoughts on whether all new verification functions should be called from `VerifyTimestampResponse` or if they should be called directly from here?

haydentherapper: I think it'd be best to have a single exported verification function that performs all other verifications. If you think there's value in exporting the sub-verification functions, maybe we put them under an internal package?

malancas: Happy to use a single exported function, I'll make that change.

pkg/verification/verify.go
```diff
@@ -18,17 +18,220 @@ package verification
...
+	Nonce          *big.Int
+	Subject        string
+	HashAlgorithm  hash.Hash
+	HashedMessage  []byte
+}
+
+func NewVerifyOpts() (VerifyOpts, error) {
+	opts := VerifyOpts{}
+
+	oidFlagVal := viper.GetString("oid")
```
haydentherapper, line 52: Sorry, I might have misunderstood, but we should keep flags out of the core library and just in the client. If you want to have this function that handles things like parsing OIDs, it should take in values that are already read in from the flags. Also, personally I prefer to not have files read from disk as part of the library, so I think this should also be handled in the client and the chain passed into the function.

So I'd say either:
* Move NewVerifyOpts to the client application with all flag parsing there
* Move NewVerifyOpts to the client app, but provide some helper utilities for constructing OIDs and splitting PEM chains into cert pools
* Keep NewVerifyOpts here, but have it take in all values (OID as a string, PEM chain as a string), and read in flags in the client app

Does that seem reasonable?

malancas: I agree about keeping the flag parsing in the `app` CLI package. Just moved that over.

cmd/timestamp-cli/app/verify.go
```diff
@@ -43,6 +47,10 @@ func addVerifyFlags(cmd *cobra.Command) {
 	cmd.MarkFlagRequired("timestamp") //nolint:errcheck
 	cmd.Flags().Var(NewFlagValue(fileFlag, ""), "cert-chain", "path to certificate chain PEM file")
 	cmd.MarkFlagRequired("cert-chain") //nolint:errcheck
+	cmd.Flags().String("nonce", "", "optional nonce passed with the request")
+	cmd.Flags().Var(NewFlagValue(oidFlag, ""), "oid", "optional oid passed with the request")
```
hectorj2f, line 51: Does this refer to the hash algorithm OID  ? If so, we might want to explain it in the description.

cmd/timestamp-cli/app/verify.go
```diff
@@ -43,6 +47,10 @@ func addVerifyFlags(cmd *cobra.Command) {
...
+	cmd.Flags().String("subject", "", "expected leaf certificate subject")
+	cmd.Flags().Var(NewFlagValue(fileFlag, ""), "cert", "path to TSA cert")
```
hectorj2f, line 53: suggestion: "path to a TSA's certificate"

cmd/timestamp-cli/app/verify.go
```diff
@@ -85,11 +93,148 @@ func runVerify() (interface{}, error) {
...
+			rootCerts = append(rootCerts, cert)
+		} else {
+			intermediateCerts = append(intermediateCerts, cert)
+		}
+		pemBytes = rest
+	}
+	return rootCerts, intermediateCerts, nil
+}
+
+func getOid() ([]int, error) {
```
hectorj2f, line 195: nit/suggestion: `getOID`

cmd/timestamp-cli/app/verify.go
```diff
@@ -85,11 +93,148 @@ func runVerify() (interface{}, error) {
...
+			return verification.VerifyOpts{}, fmt.Errorf("failed to parse cert flag value from PEM file: %w", err)
+		}
+		opts.TsaCertificate = cert
+	}
+
+	roots, intermediates, err := getCerts()
+	if err != nil {
+		return verification.VerifyOpts{}, fmt.Errorf("failed to parse root and intermediate certs from cert-chain flag: %w", err)
+	}
+	if roots != nil && intermediates != nil {
```
hectorj2f, line 130: Suggestion: The code looks good. I just share a suggestion here, feel free to adjust it. 
To avoid all these != nil if statements for all the VerifyOpts fields, you could create functions for verifyOpts as such:

`(v *VerifyOpts) OID(...)` where you set the OID if passed.
`(v *VerifyOpts) Nonce(...)` where you set the Nonce if passed. 
...
So you move all that logic inside these functions and only check for errors here.

cmd/timestamp-cli/app/verify.go
```diff
@@ -85,11 +93,148 @@ func runVerify() (interface{}, error) {
...
+		oid[i] = intVar
+	}
+
+	return oid, nil
+}
+
+func createCertFromPEMFile(certPath string) (*x509.Certificate, error) {
+	pemBytes, err := os.ReadFile(filepath.Clean(certPath))
+	if err != nil {
+		return nil, fmt.Errorf("error reading request from file: %w", err)
```
hectorj2f, line 217: nit: `error reading TSA's certificate file`

cmd/timestamp-cli/app/verify.go
```diff
@@ -85,11 +93,148 @@ func runVerify() (interface{}, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		oid[i] = intVar
+	}
+
+	return oid, nil
+}
+
+func createCertFromPEMFile(certPath string) (*x509.Certificate, error) {
```
hectorj2f, line 214: nit: I'd rename it to `parseTSACertificate(...)`, it looks like it just parses the certificate.

pkg/verification/verify.go
```diff
@@ -44,6 +166,36 @@ func VerifyTimestampResponse(tsrBytes []byte, artifact io.Reader, certPool *x509
...
+	if err != nil {
+		return err
+	}
+
+	err = verifyLeafAndIntermediatesEKU(opts)
+	if err != nil {
+		return err
+	}
+
+	err = verifyLeafCertSubject(ts.Certificates[0].Subject.String(), opts)
```
hectorj2f, line 189: Question: I am not sure the response will always contain certificates. What if the request sets `Certificates: false` ? I am just wondering it so we avoid any panic.

pkg/verification/verify.go
```diff
@@ -44,6 +166,36 @@ func VerifyTimestampResponse(tsrBytes []byte, artifact io.Reader, certPool *x509
...
+	if err != nil {
+		return err
+	}
+
+	err = verifyESSCertID(ts.Certificates[0], opts)
```
hectorj2f, line 194: nit: Perhaps we can group the ts.Certificates[0] validation into a single function with all these sub-functions.

hectorj2f (commented):
@malancas Thanks for the hard work here 👏🏻 . I added some comments and nits :).

cmd/timestamp-cli/app/verify.go
hectorj2f: If so, we could rename this flag to `hash-algorithm-oid`

malancas: No, this refers to the TSA policy OID. I'll update the description to "optional TSA policy OID passed with the request".

pkg/verification/verify.go
malancas: If the request has `Certificates: false`, maybe we could ignore the leaf cert verifications. @haydentherapper do you have any thoughts around this?

haydentherapper: If `Certificates: false`, then the leaf must be provided out of band. Either:
* You give us the certificate chain (intermediates + root), and the leaf comes from the response
* You give us the leaf + chain, and the response has no leaf
* You give us the leaf + chain, and the response has a leaf - In this case, we should just confirm that the leaves match imo

pkg/verification/verify.go
```diff
@@ -18,17 +18,175 @@ package verification
...
+	}
+
+	if errMessage != "" {
+		return errors.New(errMessage)
+	}
+	return nil
+}
+
+// Verify the leaf certificate's subject and/or subject alternative name matches a provided subject
+func verifyLeafCertSubject(cert *x509.Certificate, opts VerifyOpts) error {
```
haydentherapper, line 70: Hm, thinking more about subject verification, I'm not actually sure exactly what OpenSSL is comparing to, but I think we don't need to be as generic.

We can be more precise about what we're verifying. If you call `Subject.String`, you'll get the string representation of a pkix.Name, which will include all parts of a cert subject (common name, organization, state, etc).

I think what's most useful to compare to is the Common Name field (`cert.Subject.CommonName`). This is really only needed in the case where you aren't providing a leaf certificate and using the included cert, just to make sure you've got the expected TSA cert. I'd recommend updating to specify that we're checking the subject's Common Name only, and probably update the VerifyOpts to say CommonName rather than Subject.

pkg/verification/verify.go
```diff
@@ -18,17 +18,175 @@ package verification
...
+func verifyEmbeddedLeafCert(tsaCert *x509.Certificate, opts VerifyOpts) error {
+	if opts.TsaCertificate != nil && !opts.TsaCertificate.Equal(tsaCert) {
+		return fmt.Errorf("certificate embedded in the TSR does not match the provided TSA certificate")
+	}
+	return nil
+}
+
+func verifyLeafCert(ts timestamp.Timestamp, opts VerifyOpts) error {
+	if len(ts.Certificates) == 0 && opts.TsaCertificate == nil {
+		return fmt.Errorf("A leaf certificate must be present the in TSR or as a verify command argument")
```
haydentherapper, line 92:
```suggestion
		return fmt.Errorf("leaf certificate must be present the in TSR or as a verify option")
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,175 @@ package verification
...
+	}
+
+	errMsg := "failed to verify leaf cert"
+
+	var leafCert *x509.Certificate
+	if len(ts.Certificates) != 0 {
+		leafCert = ts.Certificates[0]
+		if opts.TsaCertificate != nil && !leafCert.Equal(opts.TsaCertificate) {
+			return fmt.Errorf("The leaf certificate included in the TSR does not match the one provied as a verify command argument")
```
haydentherapper, line 101:
```suggestion
			return fmt.Errorf("leaf certificate included in the TSR does not match the one provided as a verify option")
```

haydentherapper: nit elsewhere, error messages should start lower-cased (i don't love this as go style, but oh well haha)

pkg/verification/verify.go
```diff
@@ -18,17 +18,175 @@ package verification
...
+
+	if cert.ExtKeyUsage[0] != x509.ExtKeyUsageTimeStamping {
+		return fmt.Errorf("leaf cert EKU is not set to TimeStamping as required")
+	}
+	return nil
+}
+
+// Verify the TSA certificate and the intermediates (called "EKU chaining") all
+// have the extended key usage set to only time stamping usage
+func verifyLeafAndIntermediatesEKU(opts VerifyOpts) error {
```
haydentherapper, line 139: For only the leaf cert, you should also check the EKU is set to critical - https://github.com/sigstore/timestamp-authority/blob/main/pkg/x509/x509.go#L77-L83. Probably could check this in verifyLeafCert

(There's nothing in the RFC about criticality for intermediates)

pkg/verification/verify.go
```diff
@@ -75,3 +254,18 @@ func verifyHashedMessages(hashAlg hash.Hash, hashedMessage []byte, artifactReade
 
 	return nil
 }
+
+func CreateTimestampResponse(tsrBytes []byte) (timestamp.Timestamp, error) {
```
haydentherapper, line 258: Is this needed in verify?

haydentherapper (changes requested):
This is looking great!

pkg/verification/verify.go
malancas: Added this in [ff25635](https://github.com/sigstore/timestamp-authority/pull/153/commits/ff2563516f66753fdeab898be4c6128c8f292194)

pkg/verification/verify.go
malancas: No. It looks like the function isn't used anywhere, so I'll remove it.`VerifyTimestampResponse` will parse the bytes and check TSR status.

cmd/timestamp-cli/app/verify.go
malancas: Since the `VerifyOpts` fields will be initialized to their respective zero values when the struct instance is created, we could just set the values returned by the getter functions like `getCerts` without checking whether the values are the zero values or not. If the values are equivalent to the zero values, nothing will change. But I'm happy to add methods if you prefer.

cmd/timestamp-cli/app/verify.go
```diff
@@ -67,7 +75,7 @@ func runVerify() (interface{}, error) {
 		return nil, fmt.Errorf("error reading request from file: %w", err)
 	}
 
-	certChainPEM := viper.GetString("cert-chain")
+	certChainPEM := viper.GetString("certificate-chain")
```
haydentherapper, line 78: This and the CertPool parameter to VerifyTimestampResponse can be removed now because VerifyOpts contains the roots/intermediates

cmd/timestamp-cli/app/verify.go
```diff
@@ -85,11 +93,139 @@ func runVerify() (interface{}, error) {
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("error reading request from file: %w", err)
+	}
+
+	certs, err := cryptoutils.UnmarshalCertificatesFromPEM(pemBytes)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to parse intermediate and root certs from PEM file: %w", err)
+	}
+
+	if len(certs) < 2 {
```
haydentherapper, line 181: We should change to len(certs) == 0 or len(certs) < 1, because you can provide only a root.

pkg/tests/cli_test.go
```diff
@@ -82,17 +82,17 @@ func TestVerify(t *testing.T) {
 	tsrPath := getTimestamp(t, restapiURL, artifactContent)
 
 	// write the cert chain to a PEM file
-	pemPath := getCertChainPEM(t, restapiURL)
+	pemPath := writeCertChainToPEMFile(t, restapiURL)
 
 	// It should verify timestamp successfully.
-	out := runCli(t, "--timestamp_server", restapiURL, "verify", "--timestamp", tsrPath, "--artifact", artifactPath, "--cert-chain", pemPath)
+	out := runCli(t, "--timestamp_server", restapiURL, "verify", "--timestamp", tsrPath, "--artifact", artifactPath, "--certificate-chain", pemPath)
```
haydentherapper, line 88: Can we test all flags, nonce, OID, common-name, etc? And also a test with --certificate?

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
 	"github.com/digitorus/timestamp"
+	"github.com/pkg/errors"
 )
 
+var (
+	// EKUOID is the Extended Key Usage OID, per RFC 5280
+	EKUOID = asn1.ObjectIdentifier{2, 5, 29, 37}
+)
+
+// VerifyOpts contains verification options passed via the CLI vefify command
```
haydentherapper, line 37:
```suggestion
// VerifyOpts contains verification options for a RFC3161 timestamp
```

haydentherapper: Removed reference to CLI since this will be used by other Golang libraries

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+// These fields are then used to verify the TSR
```
haydentherapper, line 38:
```suggestion
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+type VerifyOpts struct {
+	// verifies that the TSR's OID has an expected value
```
haydentherapper, line 40:
```suggestion
	// OID verifies that the TSR's OID has an expected value
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+	OID asn1.ObjectIdentifier
+	// verifies that the TSR uses the TSACertificate as expected
```
haydentherapper, line 42:
```suggestion
	// TSACertificate verifies that the TSR uses the TSACertificate as expected
```

haydentherapper: ditto for the rest, just a nit for comment style

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+	TSACertificate *x509.Certificate
+	// verifies the TSR's certificate chain with the root certificates
```
haydentherapper, line 44:
```suggestion
	// Intermediates verifies the TSR's certificate. Optional, used for chain building
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+	Intermediates []*x509.Certificate
+	// verifies the TSR's certificate chain with the intermediate certificates
```
haydentherapper, line 46:
```suggestion
	// Roots is the set of trusted root certificates that verifies the TSR's certificate
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+	Roots []*x509.Certificate
+	// verifies that the TSR contains the expected nonce that was optionally
+	// passed to the TSA when requesting a timestamp
+	Nonce *big.Int
+	// verifies that the leaf certificate subject Common Name an expected value
```
haydentherapper, line 51:
```suggestion
	// CommonName verifies that the TSR certificate subject's Common Name matches the expected value
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+	}
+	return nil
+}
+
+// Verify the leaf's EKU is set to critical, per RFC 3161 2.3
+func verifyLeafCertEKU(cert *x509.Certificate) error {
+	var criticalEKU bool
+	for _, ext := range cert.Extensions {
+		if ext.Id.Equal(EKUOID) {
+			criticalEKU = ext.Critical
```
haydentherapper, line 97: nit: Can break once you find the extension

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+	if len(ts.Certificates) == 0 && opts.TSACertificate == nil {
+		return fmt.Errorf("leaf certificate must be present the in TSR or as a verify option")
+	}
+
+	errMsg := "failed to verify leaf cert"
+
+	var leafCert *x509.Certificate
+	if len(ts.Certificates) != 0 {
+		leafCert = ts.Certificates[0]
+		if opts.TSACertificate != nil && !leafCert.Equal(opts.TSACertificate) {
```
haydentherapper, line 116: Is this the same check you do in `verifyEmbeddedLeafCert`?

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+		return fmt.Errorf("%s: %w", errMsg, err)
+	}
+
+	return nil
+}
+
+func verifyExtendedKeyUsage(cert *x509.Certificate) error {
+	certEKULen := len(cert.ExtKeyUsage)
+	if certEKULen != 1 {
+		return fmt.Errorf("cert has %d extended key usages, expected only one", certEKULen)
```
haydentherapper, line 149:
```suggestion
		return fmt.Errorf("certificate has %d extended key usages, expected only one", certEKULen)
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+	}
+	return nil
+}
+
+func verifyLeafCert(ts timestamp.Timestamp, opts VerifyOpts) error {
+	if len(ts.Certificates) == 0 && opts.TSACertificate == nil {
+		return fmt.Errorf("leaf certificate must be present the in TSR or as a verify option")
+	}
+
+	errMsg := "failed to verify leaf cert"
```
haydentherapper, line 111:
```suggestion
	errMsg := "failed to verify TSA certificate"
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+}
+
+func verifyExtendedKeyUsage(cert *x509.Certificate) error {
+	certEKULen := len(cert.ExtKeyUsage)
+	if certEKULen != 1 {
+		return fmt.Errorf("cert has %d extended key usages, expected only one", certEKULen)
+	}
+
+	if cert.ExtKeyUsage[0] != x509.ExtKeyUsageTimeStamping {
+		return fmt.Errorf("leaf cert EKU is not set to TimeStamping as required")
```
haydentherapper, line 153:
```suggestion
		return fmt.Errorf("leaf certificate EKU is not set to TimeStamping as required")
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+	return nil
+}
+
+// Verify the TSA certificate and the intermediates (called "EKU chaining") all
+// have the extended key usage set to only time stamping usage
+func verifyLeafAndIntermediatesEKU(opts VerifyOpts) error {
+	if opts.TSACertificate == nil || opts.Intermediates == nil {
+		return nil
+	}
+	leafCert := opts.TSACertificate
```
haydentherapper, line 164: This needs to account for the leaf certificate coming from either the options or embedded in the timestamp.

pkg/verification/verify.go
```diff
@@ -18,17 +18,196 @@ package verification
...
+			criticalEKU = ext.Critical
+		}
+	}
+	if !criticalEKU {
+		return errors.New("certificate must set EKU to critical")
+	}
+	return nil
+}
+
+func verifyLeafCert(ts timestamp.Timestamp, opts VerifyOpts) error {
```
haydentherapper, line 106: Can you also add tests for verifyLeafCert, in particular that it handles getting the certificate from either the response or from the options?

pkg/verification/verify.go
malancas: Yes, it is. I'll drop this check.

pkg/verification/verify.go
malancas: I'm going remove this function and add the leaf certificate check to the larger leaf certificate verification function. The intermediate verification will be done on its own.

malancas:
@haydentherapper there are changes to the `PKCS7` methods in https://github.com/haydentherapper/pkcs7/pull/2 that can be used in this pull request. I think it would be better to merge that one first and then pull the updated dependency into this pull request. The dependency update should be straight forward.

cmd/timestamp-cli/app/verify.go
```diff
@@ -67,27 +75,143 @@ func runVerify() (interface{}, error) {
...
+			return verification.VerifyOpts{}, fmt.Errorf("failed to parse cert flag value from PEM file: %w", err)
+		}
+		opts.TSACertificate = cert
+	}
+
+	roots, intermediates, err := getRootAndIntermediateCerts()
+	if err != nil {
+		return verification.VerifyOpts{}, fmt.Errorf("failed to parse root and intermediate certs from certificate-chain flag: %w", err)
+	}
+	if roots != nil && intermediates != nil {
```
haydentherapper, line 118: It should be fine to unconditionally set the options. Also if there's only a root provided and no intermediates, will this conditional be skipped?

pkg/verification/verify.go
```diff
@@ -18,17 +18,197 @@ package verification
...
+	// TSACertificate verifies that the TSR uses the TSACertificate as expected
+	TSACertificate *x509.Certificate
+	// Intermediates verifies the TSR's certificate. Optional, used for chain building
+	Intermediates []*x509.Certificate
+	// Roots is the set of trusted root certificates that verifies the TSR's certificate
+	Roots []*x509.Certificate
+	// Nonce verifies that the TSR contains the expected nonce. Optional, used when
+	// an optional nonce was passed with a request to the TSA
+	Nonce *big.Int
+	// CommonName verifies that the TSR certificate subject's Common Name matches the expected value
```
haydentherapper, line 51:
```suggestion
	// CommonName verifies that the TSR certificate subject's Common Name matches the expected value. Optional
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,197 @@ package verification
...
+	// EKUOID is the Extended Key Usage OID, per RFC 5280
+	EKUOID = asn1.ObjectIdentifier{2, 5, 29, 37}
+)
+
+// VerifyOpts contains verification options for a RFC3161 timestamp
+type VerifyOpts struct {
+	// OID verifies that the TSR's OID has an expected value. Optional, used when
+	// an alternative OID was passed with a request to the TSA
+	OID asn1.ObjectIdentifier
+	// TSACertificate verifies that the TSR uses the TSACertificate as expected
```
haydentherapper, line 42:
```suggestion
	// TSACertificate verifies that the TSR uses the TSACertificate as expected. Optional if the TSR contains the TSA certificate
```

pkg/verification/verify.go
```diff
@@ -18,17 +18,197 @@ package verification
...
+	}
+
+	if opts.TSACertificate.SerialNumber.Cmp(tsaCert.SerialNumber) != 0 {
+		return fmt.Errorf("TSR cert serial number does not match provided TSA cert serial number")
+	}
+
+	return nil
+}
+
+// Verify the leaf certificate's subject Common Name name matches a provided Common Name
```
haydentherapper, line 72:
```suggestion
// Verify the leaf certificate's subject Common Name matches a provided Common Name
```

haydentherapper (commented):
LGTM, just a few nits

cmd/timestamp-cli/app/verify.go
malancas: Yes it will be skipped, I'll update that if statement.

haydentherapper:
Another idea, can you populate `p7.Certificates` without affecting the signature verification? you might be able to, if nothing is recomputed that would mess up signature verification

haydentherapper:
Looking over the code, the signature is verified against `p7.Content`. I think `signer.IssuerAndSerialNumber` is unconditionally set, so then if you just populate `p7.Certificates = []*x509.Certificate{leafCert}`, I think it might work?

malancas:
@haydentherapper populating `p7.Certificates = []*x509.Certificate{leafCert}` seems to work. I'll push a change with that.

haydentherapper (approved):
Awesome!!

PR #155, opened 2022-11-25
inspect: remove format flag

hectorj2f:

#### Summary

We didn't set any default format, so it is hard to know which options are available for the format.

#### Release Note

Fix: set default format value for the inspect' format flag.

### DIFF
--- a/cmd/timestamp-cli/app/inspect.go
+++ b/cmd/timestamp-cli/app/inspect.go
@@ -30,7 +30,7 @@ import (
 func addInspectFlags(cmd *cobra.Command) {
 	cmd.Flags().Var(NewFlagValue(fileFlag, ""), "timestamp", "path to timestamp response to inspect")
 	cmd.MarkFlagRequired("timestamp") //nolint:errcheck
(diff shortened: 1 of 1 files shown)


hectorj2f:
There is something else missing... pushing another commit

hectorj2f:
`timestamp-cli inspect --timestamp_server http://localhost:5555 --timestamp response.tsr` I don't think we need to set the `--format` as a required flag, especially when there is a single option.

hectorj2f:
ping @cpanato

haydentherapper:
If we don’t have any other format supported, id just say to delete the flag.

hectorj2f:
I wasn't sure, there were other plans for the format flag.

hectorj2f:
If there aren't, then let's remove it.

haydentherapper:
YAGNI, we can add it back later if we add more formatting options!

hectorj2f:
@haydentherapper I removed the flag.

haydentherapper (approved):

PR #157, opened 2022-11-28
Unit Test - pkg/verification

neilnaveen:
Signed-off-by: Neil Naveen <42328488+neilnaveen@users.noreply.github.com>

#### Summary
- included additional tests for pkg/verification for better coverage

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/pkg/verification/verify_test.go
+++ b/pkg/verification/verify_test.go
@@ -19,7 +19,6 @@ import (
 	"crypto"
 	"crypto/x509"
 	"fmt"
-	"hash"
 	"io"
 	"net/http/httptest"
 	"strings"
@@ -97,8 +96,6 @@ func Test_verifyTSRWithChain(t *testing.T) {
 	type args struct {
 		tsmessage          string
 		wantEmptyCertPool  bool
-		wantNilCertPool    bool
-		wantNilTimestamp   bool
 		wantEmptyTimestamp bool
 	}
 	viper.Set("timestamp-signer", "memory")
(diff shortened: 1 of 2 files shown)


haydentherapper (changes requested):
Given these are private functions, I don't see a need to add nil checks.

`certPool` could be nil, but I don't see a need to check in VerifyTimestampResponse if this is nil because it should not be user-controlled, and the downstream libraries can handle a nil certPool anyways.

neilnaveen:
> Given these are private functions, I don't see a need to add nil checks.
> 
> `certPool` could be nil, but I don't see a need to check in VerifyTimestampResponse if this is nil because it should not be user-controlled, and the downstream libraries can handle a nil certPool anyways.

My take on this is that the code should follow defense in depth. What if downstream libraries miss the nil check and it panics? 

I also feel that adding these additional checks does not add additional complexity. 

Thoughts?

haydentherapper:
Downstream libraries can't use private functions, so no concerns there. Panics are also more useful than errors in this context, it shows where an error occurred with a stack trace. I'd like to avoid excessive nil checks even if it doesn't add complexity, because it suggests that values could be nil when either they can't or a mistake was made.

neilnaveen:
I have removed all nil cases that I added in verify.go and verify_test.go

pkg/verification/verify_test.go
```diff
@@ -19,12 +18,15 @@ import (
...
 	"crypto"
 	"crypto/x509"
+	"fmt"
 	"io"
 	"net/http/httptest"
 	"strings"
 	"testing"
 	"time"
 
+	client2 "github.com/sigstore/timestamp-authority/pkg/generated/client"
```
haydentherapper, line 28:
```suggestion
	tsaClient "github.com/sigstore/timestamp-authority/pkg/generated/client"
```

pkg/verification/verify_test.go
```diff
@@ -118,3 +91,107 @@ func TestVerifyArtifactHashedMessages(t *testing.T) {
...
+				tsmessage: "test",
+			},
+			wantErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// creating a valid timestamp and certPool
+			respBytes, certPool, err := getTimestampData(tt.args.tsmessage, c)
```
haydentherapper, line 143: Lint error of not checking `err`

haydentherapper:
PR 181 should cover this, as it adds additional unit tests for verification

Issue #135, opened 2022-11-08
Unit test for CLI requires binary


haydentherapper:
**Description**

If the binary hasn't been rebuilt recently, then this can cause test errors. Two ideas:
* Tag these tests as e2e and provide instructions for what to do before running e2e tests (`make timestamp-cli`)
* Build the CLI while running the test - if we do this, we should build it in a temporary folder and clean up after the test finishes. I kind of prefer this, as I'd rather not have e2e tagged tests if we can avoid them

neilnaveen:
I would like to work on this.

haydentherapper:
Thanks, assigned you! I'd prefer to avoid tagging the tests if possible and go with the second approach. If you find a better way to do this also, great!

neilnaveen:
At this moment the `make test` https://github.com/sigstore/timestamp-authority/blob/06993cb6c1f322f048fab2558b167bdbe7c4ccb6/Makefile#L81 builds the CLI prior to running tests, 

>Build the CLI while running the test - if we do this, we should build it in a temporary folder and clean up after the test finishes. I kind of prefer this, as I'd rather not have e2e tagged tests if we can avoid them

Is there a reason for building CLI in a temporary folder?

haydentherapper:
I was thinking this was useful for running `go test` without `make`. The main issue is that once you run `make test`, it'll cache a binary that'll be used for subsequent runs of `go test` without rebuilding, so you might get inaccurate results (I ran into this).

neilnaveen:
Instead of building the CLI in a temporary folder, could we use a `.PHONY` target on `timestamp-cli` to solve this problem?

haydentherapper:
Would that force rebuilding? I think that's a good enough solution if so!

neilnaveen:
I have addressed this issue in this PR #159

PR #159, opened 2022-11-30
Added .PHONY target for CLI rebuilding

neilnaveen:
Signed-off-by: Neil Naveen <42328488+neilnaveen@users.noreply.github.com>

#### Summary
- #135 Added .PHONY target to timestamp-cli so that the ClI would be rebuilt every time we run make 

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ gosec:
 	$(GOBIN)/gosec ./...
 
 gen: $(GENSRC)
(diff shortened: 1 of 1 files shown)


cpanato (approved):

haydentherapper (approved):
Thanks!

PR #181, opened 2022-12-09
Verify full certificate chain

malancas:

#### Summary

Closes https://github.com/sigstore/timestamp-authority/issues/121

Verify the full certificate chain by adding intermediate certs to PKCS7 Certificates field. This will be replaced by new PKCS7 library methods soon.


haydentherapper (commented):
Can you add a test for this? I believe we already have tests to show this works, so maybe a test to show when verification fails?

malancas:
@haydentherapper just added one. Let me know if that's what you had in mind.

pkg/verification/verify_test.go
```diff
@@ -469,3 +475,107 @@ func TestVerifyExtendedKeyUsage(t *testing.T) {
...
+
+	opts := VerifyOpts{
+		Roots:         []*x509.Certificate{certChain[2]},
+		Intermediates: []*x509.Certificate{intermediate},
+	}
+
+	err = verifyTSRWithChain(ts, opts)
+	if err == nil {
+		t.Error("expected verifyTSRWithChain to fail verifying the chain")
+	}
```
haydentherapper, line 580: A few more failure modes might be useful to test:

* Leave out the intermediate, confirm it doesn't verify
* Leave out the root, confirm it doesn't verify
* Leave out the root and intermediate, confirm it doesn't verify
* Generate a timestamp without the cert (in request, `Certificates: false`) and without a leaf in the options and confirm it doesn't verify

pkg/verification/verify_test.go
```diff
@@ -469,3 +475,107 @@ func TestVerifyExtendedKeyUsage(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("failed to create signed certificate: %v", err)
+	}
+
+	opts := VerifyOpts{
+		Roots:         []*x509.Certificate{certChain[2]},
+		Intermediates: []*x509.Certificate{certChain[1]},
+	}
+
+	err = verifyTSRWithChain(ts, opts)
```
haydentherapper, line 550: One other case that would be good to test is generating a timestamp without the cert (in request, `Certificates: false`), and have the cert in `opts.TSACertificate`, and confirm it verifies

malancas: All of the above test cases should be included in https://github.com/sigstore/timestamp-authority/pull/181/commits/38a6720aeda5292c589aa040fafa24ded329f6b7

pkg/verification/verify_test.go
```diff
@@ -469,3 +475,155 @@ func TestVerifyExtendedKeyUsage(t *testing.T) {
...
+	intermediate := certChain[1]
+	root := certChain[2]
+
+	// invalidate the intermediate certificate
+	var invalidIntermediate x509.Certificate = *certChain[1]
+	invalidIntermediate.RawIssuer = nil
+	invalidIntermediate.Issuer = pkix.Name{}
+
+	type test struct {
+		ts                  *timestamp.Timestamp
```
haydentherapper, line 561: nit, can you add a name field to `test` and include that in the output, to help with debugging what is expected and which failed?

Issue #164, opened 2022-12-01
Default to NTP monitoring being on


haydentherapper:
**Description**

I think this is a useful feature and we should change the flag to default to using the provided ntp monitor configuration.

cc @kommendorkapten

kommendorkapten:
Totally agree!

neilnaveen:
> **Description**
> 
> I think this is a useful feature and we should change the flag to default to using the provided ntp monitor configuration.
> 
> cc @kommendorkapten

I would like to work on this issue. What is expected in the default config? I am referring to https://github.com/sigstore/timestamp-authority/blob/25295ed8c2b807204f802584f786d4ee7bcc9f98/pkg/ntpmonitor/config.go#L26

haydentherapper:
Thanks! We already have a default config (https://github.com/sigstore/timestamp-authority/blob/main/ntpsync.yaml).

kommendorkapten:
What I think @haydentherapper is after is to change this logic: https://github.com/sigstore/timestamp-authority/blob/main/cmd/timestamp-server/app/serve.go#L84 to instead if require explicit activation via `--ntp-monitor` we should require explicit disabling like `--disable-ntp-monitor`. 

Maybe this was clear and I just misunderstood your comment @neilnaveen?

PR #186, opened 2022-12-14
Use default NTP monitoring configuration

malancas:
#### Summary

Closes https://github.com/sigstore/timestamp-authority/issues/164

The server now uses the default NTP monitoring configuration by default. This change also enables NTP monitoring on by default. A new `--disable-ntp-monitoring` flag has been added to disable NTP monitoring.


cmd/timestamp-server/app/root.go
```diff
@@ -70,7 +70,8 @@ func init() {
 	rootCmd.PersistentFlags().String("file-signer-key-path", "", "Path to file containing PEM-encoded private key. Supported formats include PKCS#1, PKCS#8, and RFC5915 for EC")
 	rootCmd.PersistentFlags().String("file-signer-passwd", "", "Password to decrypt private key")
 	// NTP time introspection
-	rootCmd.PersistentFlags().String("ntp-monitoring", "", "Path to a file configuring ntp monitoring")
+	rootCmd.PersistentFlags().String("ntp-monitoring", "ntpsync.yaml", "Path to a file configuring ntp monitoring. Defaults to ntpsync.yaml in the root dir")
```
hectorj2f, line 73: Q: why do we default to a specific filename ? How do we get that configuration file ?

What about defaulting to some hardcoded settings and change those if the user provides a different path to a config file ?

hectorj2f: cc @haydentherapper

haydentherapper: I wouldn't expect users to configure their own list typically. It's a curated list from many different sources. Ideally we'd just use the file checked into the repo.

IIRC though, golang doesn't embed files from source in the binary, so I don't think this will work. I think you'll need to use https://pkg.go.dev/embed also.

malancas: When I run the server locally, it seems to read the file correctly. However, I can use https://pkg.go.dev/embed

haydentherapper: It might be the difference between go run and building the binary? But I'm not sure

pkg/ntpmonitor/config.go
```diff
@@ -16,12 +16,18 @@
...
+	// a blank import is recommended by the Go docs
+	// when using embed with byte slices
+	_ "embed"
 	"fmt"
 	"os"
 
 	"gopkg.in/yaml.v3"
 )
 
+//go:embed ntpsync.yaml
```
malancas, line 28: After reading the embed docs, I believe the file we want to embed needs to be in the package directory calling the embed directive or a sub directory so the file was moved to this directory.

cmd/timestamp-server/app/root.go
```diff
@@ -70,7 +70,8 @@ func init() {
 	rootCmd.PersistentFlags().String("file-signer-key-path", "", "Path to file containing PEM-encoded private key. Supported formats include PKCS#1, PKCS#8, and RFC5915 for EC")
 	rootCmd.PersistentFlags().String("file-signer-passwd", "", "Password to decrypt private key")
 	// NTP time introspection
-	rootCmd.PersistentFlags().String("ntp-monitoring", "", "Path to a file configuring ntp monitoring")
+	rootCmd.PersistentFlags().String("ntp-monitoring", "ntpsync.yaml", "Path to a file configuring ntp monitoring. Uses default ntpsync.yaml configuration if none is provided")
```
malancas, line 73: Given that the default configuration usage is handled in the `pkg/ntpmonitor` package, we don't need to include `ntpsync.yaml` as the default value but I chose to leave it here for clarity.

haydentherapper, line 73: I would probably change this to default to an empty string, and document if no string is provided, then we use the embedded configuration.

pkg/ntpmonitor/config.go
haydentherapper: Nice! This looks great

pkg/ntpmonitor/config.go
```diff
@@ -37,14 +43,20 @@ type Config struct {
 // Config object with the vales found. No sanity checking is made of the
 // loaded values.
 func LoadConfig(path string) (*Config, error) {
-	data, err := os.ReadFile(path)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read file: %s %w",
-			path, err)
+	var configData []byte
+	if path == "ntpsync.yaml" {
```
haydentherapper, line 47:
```suggestion
	if path == "" {
```

haydentherapper: Due to the change of the default flag

haydentherapper (approved):

pkg/ntpmonitor/config.go
malancas: Just tested that the server works locally with this change now.

PR #202, opened 2023-01-03
Reject requests that use SHA-1

### LINKED ISSUE #201: Reject requests that uses SHA-1.
**Description**

* RFC 5816 suggests an update to RFC 3161 by utilizing ESSCertIDv2 for signer certificate, beyond SHA-1:

> The time-stamping protocol defined in [RFC 3161](https://www.rfc-editor.org/rfc/rfc3161) [[RFC3161](https://www.rfc-editor.org/rfc/rfc3161)] requires
   that the Cryptographic Message Syntax (CMS) SignedData [[RFC5652](https://www.rfc-editor.org/rfc/rfc5652)],
   used to apply a digital signature on the time-stamp token, include a
   signed attribute that identifies the signer's certificate.
>
> This identifier only allows SHA-1 [[SHA1](https://www.rfc-editor.org/rfc/rfc5816#ref-SHA1)] to be used as the hash
   algorithm to generate the identifier value.
>
>   The mechanism used in [[RFC3161](https://www.rfc-editor.org/rfc/rfc3161)] employed ESSCertID from [RFC 2634](https://www.rfc-editor.org/rfc/rfc2634)
   [[ESS](https://www.rfc-editor.org/rfc/rfc5816#ref-ESS)].  [RFC 5035](https://www.rfc-editor.org/rfc/rfc5035) [[ESSV2](https://www.rfc-editor.org/rfc/rfc5816#ref-ESSV2)] updated ESSCertID with ESSCertIDv2 to allow
   the use of any hash algorithm.

In general, SHA-1 fell out of favor for security utilizations and recently was officially [denounced by NIST](https://www.nist.gov/news-events/news/2022/12/nist-retires-sha-1-cryptographic-algorithm).

**Suggestion**

- Support ESSCertIDv2 / RFC 5816
- Implement SHA-2/SHA-3 either as a default of a configurable opt-in.

**Related discussion**
- #2

haydentherapper:
This is an outdated algorithm that should not be used.

Fixes #201

### DIFF
--- a/pkg/verification/verify_request.go
+++ b/pkg/verification/verify_request.go
@@ -1,0 +1,30 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package verification
+
+import (
+	"crypto"
+
+	"github.com/digitorus/timestamp"
+	"github.com/pkg/errors"
+)
+
+func VerifyRequest(ts *timestamp.Request) error {
+	// only SHA-1, SHA-256, SHA-384, and SHA-512 are supported by the underlying library
+	if ts.HashAlgorithm == crypto.SHA1 {
+		return errors.New("weak hash algorithm: must be SHA-256, SHA-384, or SHA-512")
+	}
+	return nil
+}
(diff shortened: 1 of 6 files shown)


pkg/verification/verify_request.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"crypto"
+
+	"github.com/digitorus/timestamp"
+	"github.com/pkg/errors"
+)
+
+func VerifyRequest(ts *timestamp.Request) error {
+	// only SHA-1, SHA-256, SHA-384, and SHA-512 are supported by the underlying library
```
malancas, line 25:
```suggestion
	// only SHA-256, SHA-384, and SHA-512 are supported by the underlying library
```

haydentherapper: What I meant here was that we only need to handle SHA-1, since only these 4 algorithms are supported. Otherwise, we'd also need to check MD4/MD5, etc.

malancas: Ah gotcha, that makes sense.

malancas (approved):

PR #215, opened 2023-01-12
Upload SLSA provenance to existing release

haydentherapper:
This changes the behavior of the provenance generator to not create a new release, since goreleaser already does. We add two steps to download the created provenance and upload it to the existing draft release.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -64,4 +64,23 @@ jobs:
     uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.4.0
     with:
       base64-subjects: "${{ needs.release.outputs.hashes }}"
-      upload-assets: true # upload to a new release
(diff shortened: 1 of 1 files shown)


hectorj2f (approved):

cpanato (approved):
we also can disable https://github.com/sigstore/timestamp-authority/blob/main/.goreleaser.yml#L95 then this part is not needed, but i am also fine with this

thanks!

haydentherapper:
If we disable it in goreleaser, would the artifacts that are produced as part of it still be uploaded to the release created in the slsa provenance generator step? I wasn't sure, and someone pointed me to this approach. I'll try this, but happy to change it to either!

haydentherapper:
Testing https://github.com/sigstore/timestamp-authority/actions/runs/3913966044

haydentherapper:
Working as expected - Draft release is created, artifact is uploaded to the same tag

PR #220, opened 2023-01-17
Fix high code scanning vulnerabilities

malancas:

#### Summary

Fix high code scanning vulnerabilities reported [here](https://github.com/sigstore/timestamp-authority/security/code-scanning). These are all concerned with `write` permissions being available at the top level of a GHA instead of at the job level.

### DIFF
--- a/.github/workflows/auto_merge.yaml
+++ b/.github/workflows/auto_merge.yaml
@@ -3,12 +3,11 @@ name: Dependabot auto-merge
 on:
   pull_request:
 
-permissions:
-  contents: write
-  pull-requests: write
-
 jobs:
   dependabot:
+    permissions:
+      contents: write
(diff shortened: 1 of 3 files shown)


.github/workflows/auto_merge.yaml
```diff
@@ -3,12 +3,11 @@ name: Dependabot auto-merge
 on:
   pull_request:
 
-permissions:
```
haydentherapper, line 6: For top level permissions, I was told to set “permissions: read-all” as the default (in case the org sets something different, this forces read only)

cpanato: we already haver a global setting in the org config to set to read all as default, but can add it here if you want as well

haydentherapper (approved):

PR #245, opened 2023-02-10
upgrade to go 1.20.1

cpanato:
#### Summary
- upgrade to go 1.20.1

### DIFF
--- a/pkg/ntpmonitor/ntpmonitor.go
+++ b/pkg/ntpmonitor/ntpmonitor.go
@@ -18,6 +18,7 @@ package ntpmonitor
 import (
 	"errors"
 	"fmt"
+	"math/rand"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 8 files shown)


haydentherapper (approved):
Thank you!

haydentherapper:
Ah, one unit test failure, looks like you need a different way to seed rand?

haydentherapper:
Oh you probably need to update the expected test results, I guess the new implementation of random doesn't result in a the same deterministic order (which is kinda strange if true?)

cpanato:
> Ah, one unit test failure, looks like you need a different way to seed rand?

I saw, did not have time yesterday to fix it, will try this weekend

cpanato:
@haydentherapper let me know if you agree with the rand change. PTAL

cpanato:
@haydentherapper PTAL

pkg/ntpmonitor/ntpmonitor.go
```diff
@@ -151,9 +152,12 @@ func (n *NTPMonitor) Start() {
 
 	delta := time.Duration(n.cfg.MaxTimeDelta) * time.Second
 	log.Logger.Info("ntp monitoring starting")
+
+	//nolint:gosec
+	r := rand.New(rand.NewSource(time.Now().UTC().UnixNano())) // initialize local pseudorandom generator //nolint:gosec
```
haydentherapper, line 157: One downside to this is the source is not thread safe, unlike the default rand source. Unfortunately golang doesn't export this source. It's not that big of a deal though since this isn't used in such a way, but it'd be nice to get thread safety in case there's a change in the future.

What do you think about making the random generator optional in `RandomChoice`? Callers that don't pass in `rand.Rand` will call `rand.Intn`, otherwise it'll call `r.Intn`.

haydentherapper: Going to resolve so we can unblock PR

haydentherapper (commented):
@cpanato could you rebase, then the tests should pass?

cpanato:
@haydentherapper done, PTAL

haydentherapper (approved):
Thank you!

Issue #229, opened 2023-01-25
Add support for JSON format requests


malancas:
**Description**

Currently the TSA only supports requests in the DER encoded TimeStampReq format defined in [RFC-3161](https://www.ietf.org/rfc/rfc3161.txt). Let's add support for making requests  in JSON format.

This can be done with with content type request headers. If the `Content-Type` header is set to `application/timestamp-query`, the TSA will expect the request's body to be DER encoded in the TimesStampReq format. If the header is set to `application/json`, the service will expect a JSON format.

PR #247, opened 2023-02-14
Add support for making timestamp requests in JSON format

malancas:
#### Summary

Closes https://github.com/sigstore/timestamp-authority/issues/229. 

https://github.com/haydentherapper/timestamp/pull/1 must be merged first.

This supports the ability to pass timestamp requests in JSON format to the TSA.


Makefile
```diff
@@ -129,10 +130,10 @@ $(SWAGGER): $(TOOLS_DIR)/go.mod
 # help
 ##################
 
-help: # Display help
+help: ## Display help
```
malancas, line 133: This was broken. I copied the new working version of this from Fulcio's Makefile [here](https://github.com/sigstore/fulcio/blob/main/Makefile#LL152-L156C28).

pkg/tests/api_test.go
```diff
@@ -71,184 +73,274 @@ func TestGetTimestampCertChain(t *testing.T) {
...
+			t.Fatalf("unexpected policy ID")
+		}
+		// check for no extensions
+		if len(tsr.Extensions) != 0 {
+			t.Fatalf("expected 0 extensions, got %d", len(tsr.Extensions))
+		}
 	}
 }
 
 func TestGetTimestampResponseWithExtsAndOID(t *testing.T) {
```
malancas, line 193: Looking back at the timestamp-cli timestamp command, it doesn't look like it supports providing extra extensions. Since this test includes setting extra extensions, are we interested in adding a flag for extra extensions as well?

pkg/api/timestamp.go
```diff
@@ -16,24 +16,123 @@ package api
...
+	"github.com/pkg/errors"
 	ts "github.com/sigstore/timestamp-authority/pkg/generated/restapi/operations/timestamp"
 	"github.com/sigstore/timestamp-authority/pkg/verification"
 )
 
+type JSONRequest struct {
+	Artifact      string   `json:"artifact"`
+	Certificates  bool     `json:"certificates"`
+	HashAlgorithm string   `json:"hashAlgorithm"`
+	Nonce         *big.Int `json:"nonce"`
```
haydentherapper, line 41: What is the reason this is a pointer?

pkg/api/timestamp.go
```diff
@@ -16,24 +16,123 @@ package api
...
+}
+
+func requestBodyToTimestampReq(reqBytes []byte, contentType string) (*timestamp.Request, error) {
+	switch contentType {
+	case "json":
+		return ParseJSONRequest(reqBytes)
+	case "timestamp-query":
+		return timestamp.ParseRequest(reqBytes)
+	default:
+		return nil, fmt.Errorf("unsupported content type")
```
haydentherapper, line 120: Related to the above comment, it would be good to confirm if we need to default to timestamp-query or not

pkg/tests/buildTestData.go
haydentherapper, line 1: nit, file name should be build_test_data.go

pkg/api/timestamp.go
```diff
@@ -16,24 +16,123 @@ package api
...
+	Artifact      string   `json:"artifact"`
+	Certificates  bool     `json:"certificates"`
+	HashAlgorithm string   `json:"hashAlgorithm"`
+	Nonce         *big.Int `json:"nonce"`
+	TSAPolicyOID  string   `json:"tsaPolicyOID"`
+}
+
+func GetHashAlgo(algo string) (crypto.Hash, error) {
+	switch algo {
+	case "sha256":
```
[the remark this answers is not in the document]
malancas: I think we should always lowercase the input. I'll add that.

pkg/api/timestamp.go
```diff
@@ -16,24 +16,123 @@ package api
...
-	req, err := timestamp.ParseRequest(requestBytes)
+	contentType, err := getContentType(params.HTTPRequest)
+	if err != nil {
+		return handleTimestampAPIError(params, http.StatusUnsupportedMediaType, err, failedToGenerateTimestampResponse)
+	}
+
+	req, err := requestBodyToTimestampReq(requestBytes, contentType)
 	if err != nil {
 		return handleTimestampAPIError(params, http.StatusBadRequest, err, failedToGenerateTimestampResponse)
 	}
```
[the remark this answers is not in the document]
malancas: I think it makes sense to rework a bit of the parsing code and call it earlier. I'll push an update with that change soon.

cmd/timestamp-cli/app/timestamp.go
```diff
@@ -118,7 +118,12 @@ func createRequestFromFlags() ([]byte, error) {
 
 func runTimestamp() (interface{}, error) {
 	fmt.Println("Generating a new signed timestamp")
-	tsClient, err := client.GetTimestampClient(viper.GetString("timestamp_server"), client.WithUserAgent(UserAgent()))
+
+	// Set the Content-Type header to application/timestamp-query for the
+	// request that will be made to the server. Since the server accepts
+	// both application/timestamp-query and application/json as consumers for
+	// the /api/v1/timestamp endpoint, we need to specify which one we want to use
+	tsClient, err := client.GetTimestampClient(viper.GetString("timestamp_server"), client.WithUserAgent(UserAgent()), client.WithContentType(client.TimestampQueryMediaType))
```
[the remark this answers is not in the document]
malancas: On main branch, the request will be rejected if no content header is provided with HTTP status code 415.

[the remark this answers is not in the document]
haydentherapper: Perfect, thanks for checking.

pkg/tests/api_test.go
haydentherapper: Sure, do you want to file an issue for this? I think it'd also be fine to say that the average CLI user won't care about extensions and we'd also support extensions via the API.

pkg/api/timestamp.go
```diff
@@ -16,30 +16,142 @@ package api
...
 
+type JSONRequest struct {
+	Artifact      string   `json:"artifact"`
+	Certificates  bool     `json:"certificates"`
+	HashAlgorithm string   `json:"hashAlgorithm"`
+	Nonce         *big.Int `json:"nonce"`
+	TSAPolicyOID  string   `json:"tsaPolicyOID"`
+}
+
+func GetHashAlg(alg string) (crypto.Hash, string, error) {
```
haydentherapper, line 45: Private function?

pkg/api/timestamp.go
```diff
@@ -16,30 +16,142 @@ package api
...
+	case "sha512":
+		return crypto.SHA512, "", nil
+	case "sha1":
+		return 0, WeakHashAlgorithmTimestampRequest, verification.ErrWeakHashAlg
+	default:
+		return 0, failedToGenerateTimestampResponse, fmt.Errorf("unsupported hash algorithm: %s", alg)
+	}
+}
+
+func ParseJSONRequest(reqBytes []byte) (*timestamp.Request, string, error) {
```
haydentherapper, line 61: Can you add a comment if this is a public function?

pkg/api/timestamp.go
```diff
@@ -16,30 +16,142 @@ package api
...
+	// parse the DER encoded timestamp request into a timestamp.Request struct
+	tsRequest, err := timestamp.ParseRequest(tsReqBytes)
+	if err != nil {
+		return nil, failedToGenerateTimestampResponse, fmt.Errorf("failed to parse Request from Request bytes: %v", err)
+	}
+
+	return tsRequest, "", nil
+}
+
+func ParseDERRequest(reqBytes []byte) (*timestamp.Request, string, error) {
```
haydentherapper, line 107: private func?

haydentherapper (commented):
just a few nits, lgtm

haydentherapper (approved):

pkg/tests/api_test.go
malancas: Supporting extensions just through the API sounds fine to me.

PR #251, opened 2023-02-16
Update policy

malancas:
#### Summary

Removing some small sections from the policy.

### DIFF
--- a/docs/tsa-policy.md
+++ b/docs/tsa-policy.md
@@ -395,13 +395,6 @@ In particular:
   external organizations supporting the TSA services including
   the applicable policies and practices.
 
-- The TSA shall make available to subscribers and relying parties its
-  practice statement, and other relevant documentation, as necessary,
-  to assess conformance to the time-stamp policy.
-
-  NOTE 2: The TSA is not generally required to make all the details
-  of its practices public.
-
 - Maintainers of the TSA shall have final authority for approving the TSA
   practice statement and ensuring that the practices are properly implemented.
   Maintainers shall also review any changes to the TSA to confirm that they
--- a/docs/tsa-policy.md
+++ b/docs/tsa-policy.md
@@ -683,9 +683,9 @@ Media handling and security:
 
 System Planning:
 
-1. Capacity demands shall be monitored and projections of future capacity r
-   equirements made to ensure that adequate processing power
-   and storage are available.
+1. Capacity demands shall be monitored and projections of future capacity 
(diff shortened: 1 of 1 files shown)


kommendorkapten:
I reckon we should remove this section too:
https://github.com/sigstore/timestamp-authority/pull/251/files#diff-669237df67d62d453aa0fb50309ac55463d0fe7f5760b750d80981dd2805d273L424-L426

> The TSA shall make available to subscribers and relying parties its practice statement, and other relevant documentation, as necessary, to assess conformance to the time-stamp policy.
> 
> NOTE 2: The TSA is not generally required to make all the details of its practices public.

That section. 

My interpretation: 
I think it's concerning as if we open up the TSA to the public, any user would be a subscriber or relying party. Even if the TSA is not open to public use for requesting a time-stamp, any verifier is considered a relying party, and so may request the practice statement.

kommendorkapten (approved):
🚀

docs/tsa-policy.md
```diff
@@ -356,27 +351,6 @@ NOTE: It is advisable that, when obtaining a time-stamp token, the subscriber
 verifies that the time-stamp token has been correctly signed and that the
 private key used to sign the time-stamp token has not been compromised.
 
-## 6.3. Relying Party Obligations
```
haydentherapper, line 359: Should we leave this section in but remove a reference to 7.1.2? It seems like this is still useful to include guidance for a relying pary

docs/tsa-policy.md
```diff
@@ -421,17 +395,6 @@ In particular:
   external organizations supporting the TSA services including
   the applicable policies and practices.
 
-- The TSA shall make available to subscribers and relying parties its
```
haydentherapper, line 424: I think this and NOTE 2 is fine to leave - "Make available documentation, as necessary", and the note says "we might not". You could change "shall" to "may" if you want to make it clearer that it's optional.

For example, if the TSA were to log timestamps to a transparency log, that should be publicly documented for auditability.

docs/tsa-policy.md
kommendorkapten: Yeah, makes sense.

docs/tsa-policy.md
kommendorkapten: Changing it to *may* is fine!

haydentherapper (approved):

kommendorkapten (approved):

PR #287, opened 2023-03-17
CHANGELOG update for 1.0.0

haydentherapper:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.0.0
+
+1.0 release of the timestamp authority. No changes from the previous release candidate.
+
+Thank you to all contributors!
+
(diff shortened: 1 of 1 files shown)


cpanato (approved):

haydentherapper:
@malancas I think https://github.com/sigstore/timestamp-authority/pull/267 affected this. Required tests aren't running. Do you know any way around this? We can't configure per file type required tests unfortunately, so I think we might have to revert 267

malancas:
@haydentherapper oh no. There may be a way around this with more GHA rules but we can revert https://github.com/sigstore/timestamp-authority/pull/267 for now to unblock this and I can investigate whether we can use some additional GHA rules in a follow up.

malancas (approved):

PR #292, opened 2023-03-21
Add workflow trigger filters

malancas:
#### Summary

Add workflow trigger filters so the Code QL and test workflows only run when Go files or CHANGELOG.md change. The latter indicates we are cutting a new releases and all CI checks should be run in this case.


cpanato (commented):
i think is better to use the action `dorny/paths-filter` to filter the trigger because it will skip the job and then in the PR status check will mark as skipped and not as waiting for required tests.


example, then in the following steps we use the `if` to skip or run it
wdyt?

```
 - uses: dorny/paths-filter@v2
        id: filter
        with:
          filters: |
            terraform:
              - 'tests/**'
 ```

malancas:
> i think is better to use the action `dorny/paths-filter` to filter the trigger because it will skip the job and then in the PR status check will mark as skipped and not as waiting for required tests.
> 
> example, then in the following steps we use the `if` to skip or run it wdyt?
> 
> ```
>  - uses: dorny/paths-filter@v2
>         id: filter
>         with:
>           filters: |
>             terraform:
>               - 'tests/**'
> ```

Sorry for the delay in response. This sounds good. I like the idea of seeing the check as skipped instead of waiting for required tests.

malancas:
Unfortunately the `Analyze (go)` check still appears as expected because it is a step created from a matrix. I tried a few different things to get this to appear at skipped. All other checks that weren't run appear as skipped.

Issue #304, opened 2023-04-03
Signer hardcodes use of SHA-256


malancas:
**Description**

When a new crypto signer is created, the hash function is hardcoded to SHA-256. This can be found on the following lines:
- https://github.com/sigstore/timestamp-authority/blob/4121517967785c508dd57f3ea1a0d05c4692cfe9/pkg/signer/signer.go#L47
- https://github.com/sigstore/timestamp-authority/blob/4121517967785c508dd57f3ea1a0d05c4692cfe9/pkg/signer/signer.go#L42

The signer should accept a hash function configured when the server is created.

**Version**

haydentherapper:
Something to investigate - How does ed25519 work with this, since it hardcodes its own hashing algorithm?

PR #328, opened 2023-04-19
Remove signer hardcoding of SHA-256

malancas:
#### Summary

Closes https://github.com/sigstore/timestamp-authority/issues/304

Currently when signers are created, SHA-256 is always used. A KMS based key may use a different supported algorithm, like SHA-384 or SHA-512, so this hardcoding should be removed.

- Adds a new timestamp-server flag `signer-hash-function` that allows the user to specify a signer hash function. This is used in several different signer schemes
- Adds two new fetch-tsa-certs flags, `intermediate-hash-function` and `leaf-hash-function`, that allow the user to specify the leaf and intermediate key hash functions
- Updates the signer code to handle a specified hash function

### DIFF
--- a/pkg/signer/file.go
+++ b/pkg/signer/file.go
@@ -30,7 +30,7 @@ type File struct {
 	crypto.Signer
 }
 
-func NewFileSigner(keyPath, keyPass string) (*File, error) {
+func NewFileSigner(keyPath, keyPass string, hashFunc crypto.Hash) (*File, error) {
 	opaqueKey, err := pemutil.Read(keyPath, pemutil.WithPassword([]byte(keyPass)))
 	if err != nil {
 		return nil, fmt.Errorf("file: provide a valid signer, %s is not valid: %w", keyPath, err)
@@ -38,18 +38,21 @@ func NewFileSigner(keyPath, keyPass string) (*File, error) {
 	// Cannot use signature.LoadSignerVerifier because the SignerVerifier interface does not extend crypto.Signer
 	switch pk := opaqueKey.(type) {
 	case *rsa.PrivateKey:
-		signer, err := signature.LoadRSAPKCS1v15SignerVerifier(pk, crypto.SHA256)
+		signer, err := signature.LoadRSAPKCS1v15SignerVerifier(pk, hashFunc)
 		if err != nil {
 			return nil, err
 		}
 		return &File{signer}, nil
 	case *ecdsa.PrivateKey:
-		signer, err := signature.LoadECDSASignerVerifier(pk, crypto.SHA256)
+		signer, err := signature.LoadECDSASignerVerifier(pk, hashFunc)
 		if err != nil {
 			return nil, err
 		}
 		return &File{signer}, nil
 	case ed25519.PrivateKey:
+		if hashFunc != crypto.SHA512 {
+			fmt.Printf("ed25519 only supports SHA512, specified hash func is %s. Using SHA512\n", hashFunc)
+		}
 		signer, err := signature.LoadED25519SignerVerifier(pk)
 		if err != nil {
 			return nil, err
--- a/pkg/signer/file.go
+++ b/pkg/signer/file.go
@@ -51,7 +51,7 @@ func NewFileSigner(keyPath, keyPass string, hashFunc crypto.Hash) (*File, error)
 		return &File{signer}, nil
 	case ed25519.PrivateKey:
 		if hashFunc != crypto.SHA512 {
-			fmt.Printf("ed25519 only supports SHA512, specified hash func is %s. Using SHA512\n", hashFunc)
+			fmt.Printf("Ed25519 only supports SHA512, specified hash func is %s. Using hash function specified by Ed25519\n", hashFunc)
 		}
 		signer, err := signature.LoadED25519SignerVerifier(pk)
 		if err != nil {
--- a/pkg/signer/memory.go
+++ b/pkg/signer/memory.go
@@ -34,7 +34,7 @@ func getCurveFromSigner(signer crypto.Signer) (elliptic.Curve, error) {
 	// the in-memory cert chain uses ECDSA keys, so we cast the public key to an ECDSA key
 	pub, ok := signer.Public().(*ecdsa.PublicKey)
 	if !ok {
-		return nil, errors.New("signer is not an ECDSA key")
+		return nil, errors.New("casting signer's public key to an ECDSA key")
 	}
 	return pub.Curve, nil
 }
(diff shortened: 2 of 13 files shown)


pkg/signer/signer.go
```diff
@@ -36,15 +36,40 @@ const TinkScheme = "tink"
...
+	case "sha384":
+		return crypto.SHA384, elliptic.P384(), nil
+	case "sha512":
+		return crypto.SHA512, elliptic.P521(), nil
+	default:
+		return 0, nil, fmt.Errorf("invalid hash algorithm - must be either sha256, sha384, or sha512")
+	}
+}
+
+func NewCryptoSigner(ctx context.Context, signer, kmsKey, tinkKmsKey, tinkKeysetPath, hcVaultToken, fileSignerPath, fileSignerPasswd, signerHashFunc string) (crypto.Signer, error) {
```
malancas, line 52: I think this function is ready for a refactor but given that this PR already touches over ten files, it can be tackled in a follow up.

pkg/signer/file.go
```diff
@@ -30,26 +30,29 @@ type File struct {
...
 	case *ecdsa.PrivateKey:
-		signer, err := signature.LoadECDSASignerVerifier(pk, crypto.SHA256)
+		signer, err := signature.LoadECDSASignerVerifier(pk, hashFunc)
 		if err != nil {
 			return nil, err
 		}
 		return &File{signer}, nil
 	case ed25519.PrivateKey:
+		if hashFunc != crypto.SHA512 {
+			fmt.Printf("ed25519 only supports SHA512, specified hash func is %s. Using SHA512\n", hashFunc)
```
malancas, line 54: Regarding how to handle a hash function that doesn't match the one used by Ed25519, I thought it may make sense to simply inform the user starting the server that the provided algorithm will be disregarded. Thoughts?

pkg/signer/tink.go
```diff
@@ -129,6 +129,9 @@ func KeyHandleToSigner(kh *keyset.Handle) (crypto.Signer, error) {
 		p.PublicKey.X, p.PublicKey.Y = c.ScalarBaseMult(privKey.GetKeyValue())
 		return p, nil
 	case ed25519SignerTypeURL:
+		if hashFunc != crypto.SHA512 {
+			fmt.Printf("Ed25519 only supports SHA512, specified hash func is %s. Using hash function specified by Ed25519\n", hashFunc)
```
malancas, line 133: Regarding how to handle a hash function that doesn't match the one used by Ed25519, I thought it may make sense to simply inform the user that the provided algorithm will be disregarded and move on with setup. Thoughts?

pkg/signer/file.go
```diff
@@ -30,26 +30,29 @@ type File struct {
...
 	case *ecdsa.PrivateKey:
-		signer, err := signature.LoadECDSASignerVerifier(pk, crypto.SHA256)
+		signer, err := signature.LoadECDSASignerVerifier(pk, hashFunc)
 		if err != nil {
 			return nil, err
 		}
 		return &File{signer}, nil
 	case ed25519.PrivateKey:
+		if hashFunc != crypto.SHA512 {
+			fmt.Printf("Ed25519 only supports SHA512, specified hash func is %s. Using hash function specified by Ed25519\n", hashFunc)
```
haydentherapper, line 54: Should we err out instead of printing?

pkg/signer/memory.go
```diff
@@ -30,12 +30,25 @@ import (
 	tsx509 "github.com/sigstore/timestamp-authority/pkg/x509"
 )
 
+func getCurveFromSigner(signer crypto.Signer) (elliptic.Curve, error) {
```
haydentherapper, line 33: memory signer hardcodes the signer's curve - https://github.com/sigstore/timestamp-authority/blob/403a0a1fc1944a77ff4c4c3fa528f3d90c518d35/pkg/signer/signer.go#L42 - so this isn't needed

I think this is reasonable, as it's only meant for testing.

pkg/signer/signer.go
```diff
@@ -36,15 +36,40 @@ const TinkScheme = "tink"
...
 	switch signer {
 	case MemoryScheme:
-		sv, _, err := signature.NewECDSASignerVerifier(elliptic.P256(), rand.Reader, crypto.SHA256)
+		hashFunc, curve, err := getHashFuncEllipticCurve(signerHashFunc)
```
haydentherapper, line 55: I'd vote to keep this hardcoded, as it's only meant for testing and should need no additional configuration.

pkg/signer/tink.go
haydentherapper: Given this runs only on setup, I'd prefer we err out rather than allow for misconfiguration.

pkg/signer/tink.go
```diff
@@ -51,7 +51,7 @@ var (
 )
 
 // NewTinkSigner creates a signer by decrypting a local Tink keyset with a remote KMS encryption key
-func NewTinkSigner(ctx context.Context, tinkKeysetPath string, primaryKey tink.AEAD) (crypto.Signer, error) {
+func NewTinkSigner(ctx context.Context, tinkKeysetPath string, primaryKey tink.AEAD, hashFunc crypto.Hash) (crypto.Signer, error) {
```
haydentherapper, line 54: Is hash func needed here? This only builds a signer

haydentherapper (changes requested):
I think we actually need to fix this in the timestamp library, not here. sigstore/sigstore exposes a `SignMessage` interface that takes the hash passed into it, but that's only used by sigstore services. The `crypto.Signer` interface provides `Sign` which takes in a hash function in `crypto.SignerOpts`, but that's only to check that the digest provided matches the hash function expectations (like digest length). `Sign` is what's called in the [pkcs7 library](https://github.com/digitorus/pkcs7/blob/master/sign.go#L373-L404)

The hash function is passed in the timestamp library here - https://github.com/digitorus/timestamp/blob/d542479a242518f315934e8d79c8e077ddbe4990/timestamp.go#L599, and you can see it's hardcoded. I don't think any changes are needed in the pkcs7 library, as it supports other hash algorithms already - https://github.com/digitorus/pkcs7/blob/master/sign.go#L150-L156

so tl;dr - https://github.com/digitorus/timestamp/blob/d542479a242518f315934e8d79c8e077ddbe4990/timestamp.go#L599 needs to be configurable

Leaving up the comments I've typed up already, but they're moot given the above.

malancas:
Closing as the hardcoding bug should be resolved by https://github.com/sigstore/timestamp-authority/pull/452

PR #465, opened 2023-09-05
Fix bug where TSA signing fails if cert hash != content hash.

wlynch:
#### Summary

This was fixed upstream in https://github.com/digitorus/timestamp/pull/19.

We should cut a patch release - the buggy behavior was introduced in v1.1.2.

NOTE: This potentially breaks some users if they are relying on the new cert hash behavior introduced in v1.1.2 to support other hash types (since this defaults back to sha256), but this fixes those who were broken by the v1.1.2 update. Support for other hash types can be added in another PR by moving to using [timestamp.CreateResponseWithOpts](https://pkg.go.dev/github.com/digitorus/timestamp#Timestamp.CreateResponseWithOpts) and passing in the appropriate hash (but I'm considering that out of scope for this PR).

#### Release Note

- Fixes bug where signing fails if content hash != cert hash.
- BREAKING CHANGE: This reverts a new addition made in v1.1.2 that let some users use other content hashes under certain conditions but broke other users. This change reverts content hash behavior back to sha256 for the time being.

#### Documentation

I don't think we have a good spot for this at the moment? Happy to add some if you point me to the right place.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	cloud.google.com/go/security v1.15.1
 	github.com/beevik/ntp v1.3.0
 	github.com/digitorus/pkcs7 v0.0.0-20230818184609-3a137a874352
-	github.com/digitorus/timestamp v0.0.0-20230821155606-d1ad5ca9624c
+	github.com/digitorus/timestamp v0.0.0-20230902153158-687734543647
 	github.com/go-chi/chi v4.1.2+incompatible
 	github.com/go-openapi/errors v0.20.4
 	github.com/go-openapi/loads v0.21.2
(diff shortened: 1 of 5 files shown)


haydentherapper:
Can you fix the lint error @wlynch?

@malancas, did you want to add the call to CreateResponseWithOpts in a later PR?

malancas:
> Can you fix the lint error @wlynch?
> 
> @malancas, did you want to add the call to CreateResponseWithOpts in a later PR before we cut a new patch?

Yep, that sounds good. I'll take a look after this is merged.

wlynch:
> Can you fix the lint error @wlynch?

Done!

haydentherapper (approved):

PR #488, opened 2023-10-03
Support other hash algs for pre-signed timestamp besides SHA256

haydentherapper:
There are three relevant hash algs used in this codebase:

* The certificate hash alg, which specifies how the CA certificate hashed the to-be-signed certificate
* The message hash alg, specified in the request, which says how the timestamp message was hashed
* The timestamp hash alg, which specifies how the timestamp signer should hash the pre-signed timestamp structure

The latter of these three was not configurable. We had a previous approach that used the certificate hash alg, but this does not have to match the timestamp hash alg.

Fixes https://github.com/sigstore/timestamp-authority/issues/304

### DIFF
--- a/cmd/timestamp-server/app/root.go
+++ b/cmd/timestamp-server/app/root.go
@@ -59,6 +59,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&enablePprof, "enable-pprof", false, "enable pprof for profiling on port 6060")
 	rootCmd.PersistentFlags().BoolVar(&httpPingOnly, "http-ping-only", false, "serve only /ping in the http server")
 	rootCmd.PersistentFlags().String("timestamp-signer", "memory", "Timestamping authority signer. Valid options include: [kms, tink, memory, file]. Memory and file-based signers should only be used for testing")
+	rootCmd.PersistentFlags().String("timestamp-signer-hash", "sha256", "Hash algorithm used by the signer. Must match the hash algorithm specified for a KMS or Tink key. Valid options include: [sha256, sha384, sha512]. Ignored for Memory signer.")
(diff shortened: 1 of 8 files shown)


haydentherapper:
For posterity, I documented the various hashes that we see throughout the codebase to hopefully make this easier in the future. Let me know if you see any better way to do this besides hardcoding it via flag.

malancas:
> For posterity, I documented the various hashes that we see throughout the codebase to hopefully make this easier in the future. Let me know if you see any better way to do this besides hardcoding it via flag.

I had been exploring this a little while ago and it seems like we might be able to expose the signer's hash alg through the KMS clients in sigstore/sigstore. Then this codebase could just call the KMS signer's HashAlg method when it needs it. I haven't touched the code in a bit but here's a [draft PR](https://github.com/sigstore/sigstore/pull/1426) if you think it could work.

wlynch (approved):
LGTM! Also fine with the look up via API approach if/when that's ready.

haydentherapper:
@malancas What do you think about this PR for now, and then I can take a look at improving on this with your KMS PR? I'd like to cut a new release of the TSA to pick up go 1.21.3.

I like the idea of using KMS directly to provide the hash, and we can do the same with Tink since its keyset provides a hash.

malancas:
> @malancas What do you think about this PR for now, and then I can take a look at improving on this with your KMS PR? I'd like to cut a new release of the TSA to pick up go 1.21.3.
> 
> 
> 
> I like the idea of using KMS directly to provide the hash, and we can do the same with Tink since its keyset provides a hash.

That sounds good to me.

bobcallaway (approved):

PR #586, opened 2024-01-05
Use built-in feature to upload provenance to existing tag

haydentherapper:
This was failing because we upgraded to a new version of download-action while the reusable workflow uses a previous incompatible version of upload-action.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,6 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       hashes: ${{ steps.hash.outputs.hashes }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -55,6 +56,10 @@ jobs:
           set -euo pipefail
           checksum_file=$(echo "$ARTIFACTS" | jq -r '.[] | select (.type=="Checksum") | .path')
           echo "hashes=$(cat $checksum_file | base64 -w0)" >> "$GITHUB_OUTPUT"
+      
+      - name: Set tag output
```
haydentherapper, line 60: Taken from ko repo

malancas (approved):

PR #831, opened 2024-09-25
chore: bump go to use v1.23

hectorj2f:
I am bumping our go version to use 1.23 so we can solve some CVEs and upgrade our go version to latest.


cpanato (changes requested):
please also bump https://github.com/sigstore/timestamp-authority/blob/main/.github/workflows/tests.yaml#L98 to the latest minor, i think it is v1.61

cpanato (approved):

PR #927, opened 2025-01-17
chore: relax go directive to permit 1.22.x

dnwe:
#### Summary

Similar to sigstore/rekor#2323, relax the go.mod go directive to 1.22.0 here as well.

The go directive sets the minimum version of Go required to use this module. There's no need to set this to the latest semver patch version of a given Go release unless the semantics of that version of Go are required to build/test/use the module.

Ref: https://go.dev/ref/mod#go-mod-file-go

#### Release Note
NONE

#### Documentation
NONE


go.mod
```diff
@@ -37,7 +39,7 @@ require (
 	golang.org/x/net v0.34.0
 	google.golang.org/protobuf v1.36.3
 	gopkg.in/yaml.v3 v3.0.1
-	sigs.k8s.io/release-utils v0.9.0
+	sigs.k8s.io/release-utils v0.8.4
```
haydentherapper, line 42: fyi @cpanato since you maintain this - I looked into why release-utils had to bump to 1.23 and the culprit seems to be https://github.com/maxbrunsfeld/counterfeiter/, which I think could be trivially downgraded back to 1.22.

haydentherapper (approved):

go.mod
dnwe: @haydentherapper i saw this one, but an issue had already been raised and the maintainer didn’t seem keen to cut a new release before 1.24, see [here](maxbrunsfeld/counterfeiter/issues/312#issuecomment-2555606823)

dnwe: More specifically, 1.23 type aliasing had been introduced into the code so a trivial downgrade wasn’t possible

dnwe: ok, I was nerdsniped into making a PR for counterfeiter [here](/maxbrunsfeld/counterfeiter/pull/315)

haydentherapper (approved):

PR #989, opened 2025-03-12
Support self signed certificate chain

### LINKED ISSUE #988: modify certificate chain setup
@haydentherapper suggested we modify (or extend the options on) the TSA certificate chain setup (`fetch-tsa-certs`):

* current supported chain is
  * an external root certificate (on KMS CA)
  * an intermediate certificate (with key on KMS)
  * leaf cert (tinkey)
* planned cert chain is
  * self-signed root certificate (with key on KMS) 
  * leaf cert (tinkey)

The point here is that the setup becomes a little simpler (because you don't need to run a whole Certificate Authority Service in GCP for this). This is useful for sigstore specifically since the clients have the whole certificate chain anyway: a separate root does not magically make things more secure

jku:
This makes `fetch_tsa_certs.go` support a self-signed certificate chain as well:
* if `--gcp-ca-parent` is given, tool works like before (creates a chain with root, intermediate and leaf). If it is not given, the chain will contain a selfsigned root and leaf instead
* `--parent-validity=DAYS` was added since we can no longer rely on CA to set this in the self signing case
* `--intermediate-kms-resource` was renamed `--parent-kms-resource` so it works for both the intermediate case and self-signed case

Fixes #988


cmd/fetch-tsa-certs/fetch_tsa_certs.go
```diff
@@ -106,68 +120,106 @@ func fetchCertificateChain(ctx context.Context, parent, intermediateKMSKey, leaf
...
+			Subject: pkix.Name{
+				CommonName:   "sigstore-tsa-selfsigned",
+				Organization: []string{"sigstore.dev"},
+			},
+			SubjectKeyId: parentSkid,
+			NotBefore:    now,
+			NotAfter:     now.AddDate(20, 0, 0),
+			KeyUsage:     x509.KeyUsageCertSign,
+			MaxPathLen:   0,
+			IsCA:         true,
```
jku, line 148: there is something going wrong: The certificate does not contain BasicConstraints like I expect after setting these

jku: oh! `BasicConstraintsValid: true` is needed :shrug:

cmd/fetch-tsa-certs/fetch_tsa_certs.go
```diff
@@ -68,9 +80,11 @@ tinkey create-keyset --key-template ECDSA_P384 --out enc-keyset.cfg --master-key
...
 var (
-	// likely the root CA
-	gcpCaParent = flag.String("gcp-ca-parent", "", "Resource path to GCP CA Service CA")
-	// key only used for fetching intermediate certificate from root and signing leaf certificate
+	// Optional root CA
+	gcpCaRoot = flag.String("gcp-ca-parent", "", "Resource path to GCP CA Service CA. If unset, the intermediate cert will be self-signed instead")
+	// optional intermediate cert validity in days
+	intermediateValidity = flag.Int("intermediate-validity", 20*365, "Days the intermediate certificate will be valid for. Default 20*365. Value will be truncated by CA if one is used.")
+	// The kms key to use for "parent" certificate (intermediate if CA is used, self-signed certificate otherwise)
 	intermediateKMSKey = flag.String("intermediate-kms-resource", "", "Resource path to the asymmetric signing KMS key for the intermediate CA, starting with gcpkms://, awskms://, azurekms:// or hashivault://")
```
haydentherapper, line 88: Can we rename this and document that it's either the signing key for an intermediate CA when gcpCaRoot is used, or the signing key for the self-signed root CA when provided by itself?

cmd/fetch-tsa-certs/fetch_tsa_certs.go
```diff
@@ -79,18 +93,18 @@ var (
 	outputPath     = flag.String("output", "", "Path to the output file")
 )
 
-func fetchCertificateChain(ctx context.Context, parent, intermediateKMSKey, leafKMSKey, tinkKeysetPath, tinkKmsKey string,
+func fetchCertificateChain(ctx context.Context, root, parentKMSKey, leafKMSKey, tinkKeysetPath, tinkKmsKey string,
 	client *privateca.CertificateAuthorityClient) ([]*x509.Certificate, error) {
-	intermediateKMSSigner, err := kms.Get(ctx, intermediateKMSKey, crypto.SHA256)
+	parentKMSSigner, err := kms.Get(ctx, parentKMSKey, crypto.SHA256)
```
haydentherapper, line 98: Can we make the hash configurable via a flag? I believe this will err out if the hash func doesn't match what the KMS provider supports, e.g. ecdsa-p384 needs sha-384. We can default to sha256.

cmd/fetch-tsa-certs/fetch_tsa_certs.go
```diff
@@ -106,68 +120,109 @@ func fetchCertificateChain(ctx context.Context, parent, intermediateKMSKey, leaf
...
+		now := time.Now()
+		cert := &x509.Certificate{
+			SerialNumber: parentSn,
+			Subject: pkix.Name{
+				CommonName:   "sigstore-tsa-selfsigned",
+				Organization: []string{"sigstore.dev"},
+			},
+			SubjectKeyId:          parentSkid,
+			NotBefore:             now,
+			NotAfter:              now.AddDate(20, 0, 0),
```
haydentherapper, line 145: Should we use intermediateValidity (or renamed to parentValidity)?

cmd/fetch-tsa-certs/fetch_tsa_certs.go
```diff
@@ -106,68 +120,109 @@ func fetchCertificateChain(ctx context.Context, parent, intermediateKMSKey, leaf
...
+			KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+			BasicConstraintsValid: true,
+			MaxPathLen:            0,
+			MaxPathLenZero:        true,
+			IsCA:                  true,
+		}
+		ParentCertDER, err := x509.CreateCertificate(rand.Reader, cert, cert, parentPubKey, parentSigner)
```
haydentherapper, line 152:
```suggestion
		parentCertDER, err := x509.CreateCertificate(rand.Reader, cert, cert, parentPubKey, parentSigner)
```

cmd/fetch-tsa-certs/fetch_tsa_certs.go
```diff
@@ -106,68 +120,109 @@ func fetchCertificateChain(ctx context.Context, parent, intermediateKMSKey, leaf
...
+		if err != nil {
+			return nil, fmt.Errorf("creating self-signed parent certificate: %w", err)
+		}
+		parentCert, err := x509.ParseCertificate(ParentCertDER)
```
haydentherapper, line 156:
```suggestion
		parentCert, err := x509.ParseCertificate(parentCertDER)
```

haydentherapper:
Root cert LGTM!

cmd/fetch-tsa-certs/fetch_tsa_certs.go
jku: yeah I was considering replacing "intermediate" with "parent" in the CLI -- and then adding notes that it's either self-signed or intermediate

cmd/fetch-tsa-certs/fetch_tsa_certs.go
jku: yes

cmd/fetch-tsa-certs/fetch_tsa_certs.go
jku: I added a few options. The method call is a bit unwieldy now (with so many options) but I resisted urge to do more cleanup in the same PR

jku: Scratch that: I tried to do this by just providing other values to `kms.Get()` and the signature algorithm in the certificate does not change... 
* This makes sense when you consider that the (GCP) KMS key has the hash function baked in
* I'm not quite sure why `kms.Get()` requires a hash function as an argument?

In any case implementing this may not be the 15 min job I imagined: I would rather not include this in the PR. Let's file an issue if the feature is needed.

haydentherapper: Sigh, sorry for making you go down this rabbit hole. Only the [hashivault](https://github.com/sigstore/sigstore/blob/0e0ac2ead2b72d0bb713eaa126778706179b3d8a/pkg/signature/kms/hashivault/client.go#L43) KMS provider uses the hash. I thought all did to determine which hash to use when computing the digest to sign...oh well. Crypto agility is fun.

haydentherapper (approved):

PR #1018, opened 2025-03-29
expose validity period of signing cert as prometheus metric

bobcallaway:
@jku this should allow us to write an alert to monitor expiration

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,3 @@ myblob
 ts_chain.pem
 enc-keyset.cfg
 chain.crt.pem
-timestamp-cli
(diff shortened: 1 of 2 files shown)


haydentherapper (approved):
Could you add a test under https://github.com/sigstore/timestamp-authority/blob/main/pkg/tests/api_test.go as well? Also fine with doing as a follow up given we don't have tests for metrics currently

haydentherapper:
Will also need to rebase to pick up new required GHA.

haydentherapper (approved):

jku (approved):
Yes, this looks like a neat way to handle the problem. Thank you.

PR #1099, opened 2025-06-25
fix panic in cosign verify-attestation

dmitris:
#### Summary
Fix panic in `cosign verify-attestation --trusted-root trustedroot.json` due to adding a nil root certificate to the CertPool. See sigstore/cosign issue
[sigstore/cosign#4261](https://github.com/sigstore/cosign/issues/4261).

For even more robust fix, I will also do a PR in sigstore-go to avoid calling `verifyTSRWithChain` with bad / invalid `VerifyOpts`.

#### Verification
* When running the new unit test against the trunk version, getting "red" - the expected failure [gist](https://gist.github.com/dmitris/889fca18dba0a69aab9f6ca761b61e65).

* confirmed that when building `cosign` using the version of this PR (with `replace github.com/sigstore/timestamp-authority => ../timestamp-authority` in sigstore/cosign/go.mod), I no longer get a panic, the error is "regular":
```
Error: no matching attestations: failed to verify timestamps: threshold not met for verified signed timestamps: 0 < 1; error: unable to verify signed timestamps: error while verifying with chain: pkcs7: failed to verify certificate chain: x509: certificate signed by unknown authority
no valid root certificates provided for verifying the certificate chain - nil opts.Roots
error during command execution: no matching attestations: failed to verify timestamps: threshold not met for verified signed timestamps: 0 < 1; error: unable to verify signed timestamps: error while verifying with chain: pkcs7: failed to verify certificate chain: x509: certificate signed by unknown authority
no valid root certificates provided for verifying the certificate chain - nil opts.Roots
```

#### Release Note
* Bug fixes and fixes of previous known issues
Fix panic in `cosign verify-attestation` when using a trustedroot file.

#### Documentation
N/A

### DIFF
--- a/pkg/verification/verify.go
+++ b/pkg/verification/verify.go
@@ -285,7 +285,16 @@ func verifyTSRWithChain(ts *timestamp.Timestamp, opts VerifyOpts) error {
 	if len(opts.Roots) == 0 {
 		return fmt.Errorf("no root certificates provided for verifying the certificate chain")
 	}
-
+	rootFound := false
+	for _, cert := range opts.Roots {
+		if cert != nil {
+			rootFound = true
+			break
+		}
+	}
+	if !rootFound {
+		return fmt.Errorf("no valid root certificates provided for verifying the certificate chain - nil opts.Roots")
+	}
 	rootCertPool := x509.NewCertPool()
 	for _, cert := range opts.Roots {
 		rootCertPool.AddCert(cert)
--- a/pkg/verification/verify.go
+++ b/pkg/verification/verify.go
@@ -285,24 +285,18 @@ func verifyTSRWithChain(ts *timestamp.Timestamp, opts VerifyOpts) error {
 	if len(opts.Roots) == 0 {
 		return fmt.Errorf("no root certificates provided for verifying the certificate chain")
 	}
-	rootFound := false
-	for _, cert := range opts.Roots {
-		if cert != nil {
-			rootFound = true
-			break
-		}
-	}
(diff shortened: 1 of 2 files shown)


pkg/verification/verify.go
```diff
@@ -285,7 +285,16 @@ func verifyTSRWithChain(ts *timestamp.Timestamp, opts VerifyOpts) error {
...
+			rootFound = true
+			break
+		}
+	}
+	if !rootFound {
+		return fmt.Errorf("no valid root certificates provided for verifying the certificate chain - nil opts.Roots")
+	}
 	rootCertPool := x509.NewCertPool()
 	for _, cert := range opts.Roots {
 		rootCertPool.AddCert(cert)
```
bobcallaway, line 300: why not just do a check here to only add the cert to the pool if it is non-nil?

dmitris: but then how we can flag that there are no valid root certificates provided - by comparing `rootCertPool` to an empty pool, something like (using https://pkg.go.dev/crypto/x509#CertPool.Equal)?
```go
if rootCertPool.Equal(x509.NewCertPool()) {
    return fmt.Errorf("no valid root certificates provided for verifying the certificate chain")
}
```

Changed to that in 7ca231502e5bbfe05ea91744e420cefb43886361 - add to the pool only if non-empty and check for `rootCertPool` not Equal to an empty one.

bobcallaway (approved):
LGTM

haydentherapper (approved):
Thanks!

Issue #1141, opened 2025-08-20
Server reporting cert doesn't have timestamping EKU when it does, won't start


delfuego:
I'm still on my journey to get a timestamp server cert from our organizational CA, and was provided a cert this morning that *should* work, but when I start the server, it's panicking and reporting:

```
2025-08-20T14:19:00.750Z	PANIC	api/api.go:108	certificate must have extended key usage timestamping set to sign timestamping certificates
```

The cert definitely does have the timestamping EKU, though, and it's set to critical (and I'm familiar with the different error when it's not set to critical, that was one of the iterations that our PKI folks accidentally generated and sent to me previously).

The leaf/server cert that I'm using is:

```
-----BEGIN CERTIFICATE-----
MIIIBzCCBu+gAwIBAgITLQAABPpI63HNEwgf4AABAAAE+jANBgkqhkiG9w0BAQsF
ADBUMRUwEwYKCZImiZPyLGQBGRYFbG9jYWwxHzAdBgoJkiaJk/IsZAEZFg9ESEhT
U2VjdXJpdHlEZXYxGjAYBgNVBAMTEU5JSERFVi1EUEtJLVNVQkNBMB4XDTI1MDgy
MDEzNDk0NFoXDTI2MDgyMDEzNDk0NFowgeExCzAJBgNVBAYTAlVTMREwDwYDVQQI
EwhNYXJ5bGFuZDERMA8GA1UEBxMIQmV0aGVzZGExIjAgBgNVBAoTGU5hdGlvbmFs
IENhbmNlciBJbnN0aXR1dGUxPjA8BgNVBAsMNUNDUiBPZmZpY2Ugb2YgU2NpZW5j
ZSAmIENsaW5pY2FsIFJlc2VhcmNoIEluZm9ybWF0aWNzMScwJQYDVQQDEx5OQ0kg
Q0NSIE9TQ1JJIFRpbWVzdGFtcCBTZXJ2ZXIxHzAdBgkqhkiG9w0BCQEWEGxldmlu
ZWphQG5paC5nb3YwggIiMA0GCSqGSIb3DQEBAQUAA4ICDwAwggIKAoICAQDPDcJM
noII50lHvZDa+kzRZNyi1MkfvQNb99+1A1DRlCP0SJx2PLIS7dz72OJykIgY8tso
aVJpYbO9EhTatSynNk8/ysfxB10sOHfdK89zgj4Glk0SZ51xbir8BfQdzwPSSCir
[...]

PR #1146, opened 2025-08-21
add feature to disable intermediate cert EKU enforcement

delfuego:
add server command-line parameter to disable enforcement of requirement that intermediate certificates have the critical timestamping EKU present

Resolves: #1141

#### Summary

`timestamp-server` currently enforces any intermediate certificates also having the critical timestamping EKU present; this isn't required by RFC3161, and it looks like many of the online free/commercial timestamping providers don't adhere to this usage (e.g., Apple, Globalsign, Digicert). This commit adds a command-line option to disable this behavior, so that a server can use intermediate certificates without the critical timestamping EKU.

#### Release Notes

* Adds `--enforce-intermediate-eku` boolean command-line option, with a default of true, so the present behavior continues unless a server is started with the option present and set to `false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,10 @@ To deploy to production, the timestamp authority currently supports signing with
 a certificate chain (leaf, any intermediates, and root), where the certificate chain's purpose (extended key usage) is
 for timestamping. We do not recommend the file signer for production since the signing key will only be password protected.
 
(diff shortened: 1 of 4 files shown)


haydentherapper (approved):
Thanks for making this change, and thanks for investigating this!

PR #1333, opened 2026-04-13
Ensure correct certificate is used for TSA auth checks (GHSA-xm5m-wgh2-rrg3)

jku:
verifyTSRWithChain and verifyLeafCert both need to find the signer (TSA) certificate from the PKCS7 certificate bag. Currently, if the bag has been injected with another leaf certificate, verifyLeafCert may use the wrong certificate to do authorization checks on.

* Make verifyTSRWithChain return the correct signer certificate
* modify verifyLeafCert so it uses the signer certificate chosen by verifyTSRWithChain
* Refactor so that verifyTSRWithChain does two checks that were previously done later -- this guarantees that verifyTSRWithChain always returns the correct certificate or errors:
  * ensure that at an embedded certificate is found or one is provided as option (or both)
  * if a certificate is provided as option and there is an embedded certificate, ensure the two match
* Add test with TSR with injected cert

--- 

(this is identical to #1330, I'm making this PR so it doesn't look like hayden is reviewing his own changes)


Hayden-IO (approved):
Thanks for doing this fix!

PR #1375, opened 2026-05-26
Fix spec violations in policy, EKU, and hash verification

Hayden-IO:
To prevent clients from specifying their own TSA policy OIDs, obtaining signatures over arbitrary extensions, and utilizing weak digests, this change:

* Introduces server flags for default/accepted policy OIDs and custom extensions
* Configures server-side validation to reject unauthorized policy OIDs and custom request extensions by default
* Updates client-side verification to build paths including TSR-embedded intermediates and enforce EKU chaining constraints on every intermediate CA
* Configures client-side verification to reject SHA-1, which is the server already does

### DIFF
--- a/pkg/verification/verify.go
+++ b/pkg/verification/verify.go
@@ -17,6 +17,7 @@ package verification
 
 import (
 	"bytes"
+	"crypto"
 	"crypto/x509"
 	"encoding/asn1"
 	"fmt"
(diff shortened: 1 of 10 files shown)


pkg/verification/verify.go
```diff
@@ -297,21 +325,26 @@ func verifyTSRWithChain(ts *timestamp.Timestamp, opts VerifyOpts) (*x509.Certifi
 		p7Message.Certificates = []*x509.Certificate{opts.TSACertificate}
 	}
 
-	err = p7Message.VerifyWithOpts(x509Opts)
+	err = p7Message.Verify()
```
Hayden-IO, line 328: This is an intentional change - otherwise we're doing chain building twice.

bobcallaway (approved):

PR #1384, opened 2026-06-02
Validate TSA certificate chain at timestamp issuance time

codysoyland:
## Summary

- Validate the TSA certificate chain at the timestamp's `genTime` rather than at the current wall-clock time, so that verification succeeds even after the TSA certificate has expired
- Add `VerifyOpts.CurrentTime` to allow callers to override the validation time explicitly

## Motivation

Previously, `VerifyTimestampResponse` validated the TSA certificate chain using `time.Now()` (Go's default when `x509.VerifyOptions.CurrentTime` is zero). This caused verification to fail once the TSA signing certificate expired, even though the timestamp was issued while the certificate was still valid.

Per [RFC 3161](https://www.rfc-editor.org/rfc/rfc3161) Appendix B (steps B.4–B.5), the TSA certificate chain should be validated at the time the timestamp was issued (`genTime`), not at the time of verification. A timestamp token remains valid as long as the TSA certificate was valid and unrevoked at `genTime`.

This change defaults `CurrentTime` to `ts.Time` when the caller does not specify one, aligning certificate chain validation with the RFC 3161 verification model.

### DIFF
--- a/pkg/verification/verify.go
+++ b/pkg/verification/verify.go
@@ -24,6 +24,7 @@ import (
 	"hash"
 	"io"
 	"math/big"
+	"time"
 
 	"github.com/digitorus/pkcs7"
 	"github.com/digitorus/timestamp"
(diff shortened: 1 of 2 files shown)


Hayden-IO (approved):
To confirm, was there a regression or was this a new conformance test recently added?

codysoyland:
> To confirm, was there a regression or was this a new conformance test recently added?

As we discussed in Slack, this is a regression due to the underlying pkcs7 library's defaults when we switched from `VerifyWithOpts` to `Verify` in https://github.com/sigstore/timestamp-authority/pull/1375. -- `VerifyWithOpts` set the CurrentTime to the signing time (if present), but `Verify` uses the current time.

PR #1389, opened 2026-06-08
count unknown EKUs when enforcing single timestamping usage

dxbjavid:
the leaf eku check enforces rfc 3161 2.3's single timestamping usage by counting cert.ExtKeyUsage, but crypto/x509 records key purposes it doesn't recognise under UnknownExtKeyUsage, so a tsa leaf carrying id-kp-timeStamping plus an unknown eku oid is accepted as a timestamping certificate both when verifying a response and when validating the configured chain. that lets a broader multi-purpose certificate stand in where the spec wants a timestamping-only one. count the known and unknown key purposes together at both sites so only a leaf with exactly the one usage is allowed.

### DIFF
--- a/pkg/verification/verify.go
+++ b/pkg/verification/verify.go
@@ -137,12 +137,15 @@ func verifyLeafCert(leafCert *x509.Certificate, verifiedChains [][]*x509.Certifi
 }
 
 func verifyLeafExtendedKeyUsage(cert *x509.Certificate) error {
-	certEKULen := len(cert.ExtKeyUsage)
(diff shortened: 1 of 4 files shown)


Hayden-IO (commented):
Thanks for fixing this, good catch! Can you please sign the DCO?

dxbjavid:
done, signed off and force-pushed. should be clean now.

Hayden-IO (approved):
Thank you again for fixing this!

dxbjavid:
Thank you for the review and merge. Much appreciated

PR #1402, opened 2026-06-17
bound json nonce length before parsing in parsejsonrequest

dxbjavid:
**Quadratic nonce parsing on the JSON timestamp endpoint**

The JSON branch of `/api/v1/timestamp` decodes the request's nonce straight into a big.Int, and base-10 integer parsing is quadratic in the digit count. Because the endpoint is reachable before any authentication, a request carrying a long decimal nonce (comfortably inside the default 1MB body limit) ties up a fair amount of CPU: a 200k-digit nonce already costs roughly 30ms and the cost grows with the square of the length, so a body near the limit reaches several hundred milliseconds per request. The timestamp-query (DER) path is not affected because ASN.1 integers are read from their binary form, which is linear.

The artifact hash, hash algorithm and policy OID are all small fixed fields, so the only thing that can grow without limit is the nonce. Rather than singling out the nonce, the change caps the whole JSON request at 1KB before parsing, which leaves ample room for a real request and rejects the oversized case up front.

### DIFF
--- a/pkg/api/timestamp.go
+++ b/pkg/api/timestamp.go
@@ -42,13 +42,15 @@ type JSONRequest struct {
 	TSAPolicyOID  string   `json:"tsaPolicyOID"`
 }
 
-// maxJSONNonceBytes bounds the length of the JSON-encoded nonce before it is
-// decoded into a big.Int. RFC 3161 nonces are small random values, but parsing
-// a base-10 integer is quadratic in the number of digits, so an unbounded
-// decimal nonce lets a single request burn a disproportionate amount of CPU.
-// The DER request path is unaffected because ASN.1 INTEGERs are decoded from
-// their binary representation.
-const maxJSONNonceBytes = 1024
+// maxJSONRequestBytes bounds the size of a JSON timestamp request before it is
+// parsed. The artifact hash, hash algorithm and policy OID are all small fixed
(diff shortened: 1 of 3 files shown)


pkg/api/timestamp.go
```diff
@@ -60,6 +68,20 @@ func getHashAlg(alg string) (crypto.Hash, string, error) {
 
 // ParseJSONRequest parses a JSON request into a timestamp.Request struct
 func ParseJSONRequest(reqBytes []byte) (*timestamp.Request, string, error) {
+	// Bound the nonce length before it is decoded into a big.Int. The raw
```
Hayden-IO: Could we put a reasonable limit in place for the whole request, rather than just the nonce? `TSAPolicyOID`, `HashAlgorithm`, `ArtifactHash` all have a max size, so we can tack on a little more for the nonce. 1Kb for the entire request?

dxbjavid: makes sense, that's cleaner. switched it to cap the whole request at 1KB up front and dropped the nonce-specific probe. the other fields are all small and fixed so a single bound covers it, and the oversized nonce never reaches the big.Int decode. updated the test to match.

Hayden-IO (approved):

Hayden-IO:
Thanks!

dxbjavid:
Thank you for the review and merge. Appreciated

PR #1405, opened 2026-06-21
reject timestamp response missing the requested nonce

dxbjavid:
verifyNonce compares the caller's expected nonce against the nonce parsed from the response, but the nonce is optional in an RFC 3161 reply, so a validly signed response that omits it leaves the parsed value nil and the comparison panics once a caller has requested a nonce. It is reachable after the signature and chain have already verified, for instance when an older timestamp carrying no nonce is replayed for the same artefact. Reject a response that lacks the expected nonce rather than dereferencing it, which also lines up with RFC 3161 2.4.2 requiring a requested nonce to be echoed back.

### DIFF
--- a/pkg/verification/verify.go
+++ b/pkg/verification/verify.go
@@ -227,6 +227,14 @@ func verifyNonce(requestNonce *big.Int, opts VerifyOpts) error {
 	if opts.Nonce == nil {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


Hayden-IO (approved):
Thank you!

dxbjavid:
Thank you for reviewing and merging this PR.

PR #1416, opened 2026-06-30
require timestamping leaf certificate to be an end-entity

dxbjavid:
verifyLeafCert and VerifyCertChain enforce that a timestamping leaf has a single critical timestamping EKU but never look at its basic constraints, so a certificate that asserts the CA bit while carrying that EKU is accepted as a TSA signer. crypto/x509 chain verification does not reject a CA certificate used as the leaf, so nothing else catches it, and that goes against RFC 3161 2.3 which expects the TSA certificate to be an end-entity and keeps certificate-issuing keys separate from timestamp-signing keys. This rejects a leaf whose basic constraints mark it as a CA in both the response-verification and issuance-chain paths.

### DIFF
--- a/pkg/verification/verify.go
+++ b/pkg/verification/verify.go
@@ -111,6 +111,14 @@ func verifyLeafCert(leafCert *x509.Certificate, verifiedChains [][]*x509.Certifi
 
 	errMsg := "failed to verify TSA certificate"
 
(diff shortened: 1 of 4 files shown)


Hayden-IO (approved):
Thank you very much!

dxbjavid:
I'm glad to hear that. Thank you very much for the review and approval! I really appreciate your time and feedback.

PR #1429, opened 2026-07-18
fail startup when the server config file cannot be read

dxbjavid:
initConfig throws away the error from viper.ReadInConfig, so a config file that exists but is malformed or unreadable is silently ignored, and that includes one named explicitly with --config. Every setting then falls back to its flag default, and the default for timestamp-signer is memory, so NewAPI takes the in-memory branch and generates a throwaway self-signed chain at startup rather than using the configured KMS, Tink or file signer; the server comes up healthy and issues timestamps under that ephemeral key, with accepted-policy-oids and enforce-intermediate-eku quietly reverted too. I reproduced it by putting a YAML syntax error in a timestamp-server.yaml that selects the file signer: the server started with no complaint and /api/v1/timestamp/certchain served the built-in "Test TSA Root" chain. This now treats only ConfigFileNotFoundError as benign and makes any other read or parse error fatal, which is what timestamp-cli already does with the same call, so a config path that is given but cannot be read will fail at startup instead of being skipped.

### DIFF
--- a/cmd/timestamp-server/app/root.go
+++ b/cmd/timestamp-server/app/root.go
@@ -16,6 +16,7 @@
 package app
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"time"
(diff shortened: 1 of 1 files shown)


Hayden-IO (approved):
Thanks!

dxbjavid:
TY!

PR #1442, opened 2026-07-29
allow crypto/x509 to generate serial number and skid

rolandshoemaker:
#### Summary

crypto/x509.CreateCertificate will generate a compliant serial number and SKID itself if one is not set in the template, no need to duplicate that functionality. Also, side-note, SKID in the leaf is redundant, and is only necessary in certificates with a basic constraints extension with cA: true (see https://tools.ietf.org/html/rfc5280#section-4.2.1.2).

#### Release Note

NONE

#### Documentation

This change is opaque to users.

### DIFF
--- a/pkg/signer/memory.go
+++ b/pkg/signer/memory.go
@@ -26,7 +26,6 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/sigstore/sigstore/pkg/cryptoutils"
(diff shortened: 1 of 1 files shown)


Hayden-IO (commented):
Thanks Roland for the cleanup, can you sign the DCO and fix the lint error? I'm also happy to take over the PR too.

Hayden-IO (approved):

Hayden-IO:
Related cleanup - https://github.com/sigstore/fulcio/pull/2416

PR #1453, opened 2026-08-11
Update release pipeline to match sigstore/rekor-tiles

Hayden-IO:
This change begins the process of unifying the build and release pipelines for Sigstore services. We start with the TSA, because it's already building on GH rather than GCB, and because the slsa-github-generator project is no longer maintained so we need to update how we generate provenance.

This includes the following changes:

* Publish provenance using actions/attest for containers and binaries, which will be documented in release notes. This also drops using Cosign, as the attest action will generate signed provenance and upload it to the registry as a referring artifact.
* Switch base image to distroless/static-debian13, and switch to defining the containers with a multi-stage dockerfile
* When building container images, build Go binaries with the same architecture to speed up build
* Drop arm7 architecture, pick a consistent set of architectures for servers and clients
* Update linting to use zizmor and a consistent zizmor and golangci-lint version


Hayden-IO:
@loosebazooka @jku I know you haven't made changes in this repo, but wanted another set of eyes to double check I haven't done something wrong in copying how we build rekor-tiles.

bobcallaway (dismissed):

jku (approved):

jku:
Looks right to me

Hayden-IO:
Will get https://github.com/sigstore/rekor-tiles/pull/926 in first, and then update this PR to match that pattern.

Dockerfile
```diff
@@ -25,21 +28,53 @@ ADD ./cmd/ $APP_ROOT/src/cmd/
...
-FROM golang:1.27.1@sha256:f44f6e88636cfb311f9ebace870ded69d943f227bb3cb27d32ffd84ea18c43ea AS deploy
+# Build server for deployment
+RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} CGO_ENABLED=0 go build -ldflags "${SERVER_LDFLAGS}" -o timestamp-server ./cmd/timestamp-server
 
+# Production deployment build
+FROM --platform=$BUILDPLATFORM gcr.io/distroless/static-debian13:nonroot@sha256:e2e927ec666bae08560abb3c55d0659eceabb657f56b6782ab500a9fc7f555e3 AS deploy
 # Retrieve the binary from the previous stage
 COPY --from=builder /opt/app-root/src/timestamp-server /usr/local/bin/timestamp-server
+# sleep needed for graceful shutdown
+COPY --from=builder /usr/bin/sleep /usr/bin/sleep
```
loosebazooka, line 39: this is a bit confusing, the builder image isn't multi platform is it?

Hayden-IO: My understanding is it is - https://github.com/docker-library/golang/blob/c4664da1bd8d0cbc975460744d90c031b409c215/1.27/trixie/Dockerfile - just based on it pulling different Go installations. Do you know of a way to verify this?

loosebazooka: well if BUILDPLATFORM is different than TARGETARCH then we end up in a weird state where sleep is pulled from the BUILDPLATFORM image but not necessarily compatible with the TARGETARCH. Perhaps this isn't an issue because we always run on amd64 + we don't see these sleep issues.

Additionally it seems like line 35 should select based on the TARGETARCH? as well?

loosebazooka: I actually think we need to fix this in both builds for "correctness", though the current state is serviceable.

Rekor tiles hasn't published one of these yet, but at head if we do a build targeting arm64 from a amd64 machine you end up with the amd64 sleep util on the arm64 image. Further more, thought the distroless/static image is so very minimal there isn't much architecture specific, there may be image metadata that is incorrectly representing the target platform of our builds.

Dockerfile
```diff
@@ -25,21 +28,53 @@ ADD ./cmd/ $APP_ROOT/src/cmd/
...
-RUN go build -ldflags "${SERVER_LDFLAGS}" ./cmd/timestamp-server
-RUN CGO_ENABLED=0 go build -gcflags "all=-N -l" -ldflags "${SERVER_LDFLAGS}" -o timestamp-server_debug ./cmd/timestamp-server
-
-# Multi-Stage production build
-FROM golang:1.27.1@sha256:f44f6e88636cfb311f9ebace870ded69d943f227bb3cb27d32ffd84ea18c43ea AS deploy
+# Build server for deployment
+RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} CGO_ENABLED=0 go build -ldflags "${SERVER_LDFLAGS}" -o timestamp-server ./cmd/timestamp-server
 
+# Production deployment build
+FROM --platform=$BUILDPLATFORM gcr.io/distroless/static-debian13:nonroot@sha256:e2e927ec666bae08560abb3c55d0659eceabb657f56b6782ab500a9fc7f555e3 AS deploy
```
jku, line 35: @loosebazooka mentioned this but I'll point it out here: BUILDPLATFORM here and in other stages seems wrong. You should be able to just remove `--platform` from all of these since (https://docs.docker.com/reference/dockerfile/#from): 

> By default, the target platform of the build request is used

Dockerfile
jku: copying /usr/bin/sleep might be a bad idea to begin with: it's coming from debian so is likely dynamically linked... but we expect it to run on distroless (with no glibc I assume?)

I assume sleep is for k8s reasons? I've seen people compile a tiny go code snippet for this so that might be an option if it's really needed.

jku:
should we add a smoketest for the arm container builds? It looks like if we don't bugs are likely...

Hayden-IO:
@jku Added that, thanks for the idea, https://github.com/sigstore/timestamp-authority/pull/1453/changes#diff-95557d53bf91069e59d82b9d8fcfaf52ec84a762858983edd5ef3c9b3e4c8191R47-R91

PR #1458, opened 2026-08-19
parse content-type header as a media type

dxbjavid:
getContentType routes a timestamp request by counting and splitting the Content-Type header on the substring application/, so a standard header like application/json; charset=utf-8 is refused while an unrelated type such as multipart/form-data; boundary=application/json is accepted and handed to the JSON parser because its boundary parameter contains the substring. Both come from matching the raw header text rather than the actual media type on an unauthenticated request. Parse the header with mime.ParseMediaType and match the media type against application/ exactly, which drops any parameters before the content-type switch and rejects anything that is not really an application type.

### DIFF
--- a/pkg/api/timestamp.go
+++ b/pkg/api/timestamp.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"mime"
 	"net/http"
 	"slices"
 	"strings"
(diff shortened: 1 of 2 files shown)


Hayden-IO (approved):
thank you, great catch!

dxbjavid:
Appreciated. Thanks!

PR #1472, opened 2026-09-01
feat: add support for ML-DSA

### LINKED ISSUE #1471: ML-DSA support
Placeholder for support for ML-DSA timestamps and supporting cryptographic structures

1. requirements
1. simple design for work involved
1. actual work

knrc:
#### Summary
This PR adds support for ML-DSA.  Resolves #1471.

#### Release Note
Added support for using ML-DSA keys and certificate chains


knrc:
Note: this is in draft mode as it's tracking Hayden's sigstore PR

pkg/signer/file_test.go
```diff
@@ -66,6 +69,21 @@ func TestNewFileSigner(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// TODO - cryptoutils' encrypted-PEM marshalling does not yet support ML-DSA.
```
jku: can you expand on this? I don't quite understand the reference to encryption but I would have expected `MarshalPrivateKeyToPEM(mldsaKey)` to work

knrc: You are right, I switched back to the encrypted functions and it does work.  I'll update the PR.  I obviously made a mistake the first time around, and should have dug into it.

knrc: I've switched back to the encrypted routines and updated the PR

knrc:
@bobcallaway @steiza rebased and updated to use sigstore main instead of Hayden's PR

pkg/api/api.go
```diff
@@ -79,6 +80,10 @@ func NewAPI() (*API, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "getting new tsa signer")
 	}
+	if _, ok := tsaSigner.Public().(*mldsa.PublicKey); ok {
+		// pkcs7 forces SHA-512 for ML-DSA SignerInfo digests.
```
steiza, line 84: Do you have some links that provide more information about this claim?

In general, [Design: Post-Quantum Crypto for Sigstore](https://docs.google.com/document/d/1g3WQk4JSwQ26MGwB-Kkhu8ylrPF5JnE4v62d2tDC2xY/edit?usp=sharing) uses SHA256. There was some discussion about opinionated selection of hash algorithms on https://github.com/sigstore/protobuf-specs/pull/898/changes as well.

In addition, https://github.com/sigstore/protobuf-specs/blob/d615235697f53603ef41ebf6e582582e4e0ddd90/protos/sigstore_common.proto#L167-L171 just contains the bytes, without room to specify the hash algorithm.

[FIPS 204](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf) supports signing and verifying with SHA256, but admittedly there are gaps in my knowledge here, so references are appreciated!

knrc: It reflects the current implementation within the pkcs7 code, which could still be a work in progress.  That code always hardcodes the digest to SHA-512.  The RFC allows for others digests, dependent on the algorithm, with SHA-512 being the interoperable setting.

ML-DSA-44 - SHA-256, SHA-384, SHA-512, SHA3-256, SHA3-384, SHA3-512, SHAKE128, SHAKE256
ML-DSA-65 - SHA-384, SHA-512, SHA3-384, SHA3-512, SHAKE256
ML-DSA-87 - SHA-512, SHA3-512, SHAKE256

jku: Yeah quick look at the rfc and code seems to support this:
* RFC9882 (ML-DSA in CMS) says 
  * SHA-512 MUST be supported
  * other algorithms MAY be supported (with the limitations in previous comment)
* digitorus/pkcs7 only supports SHA-512 when signing: this limitation is explicitly mentioned in a comment and the test suite tests that the provided `digestOID` argument is **not** used -- but of course it's possible this is just waiting for an improvement to arrive

So with current pkcs7, we will be using ML-DSA with SHA512 only.

That said, I don't think we should just drop the `timestamp-signer-hash` argument... 
* At the very least if someone explicitly sets `timestamp-signer-hash` to a value that's incompatible, we should fail
* *Maybe*  there's a case to be made that the default `timestamp-signer-hash` should change to SHA-512 when the signing key is ML-DSA... unsure about this one

steiza: Aha, thank you both for that context! With that in mind, why don't we:

- move this check into `pkg/signer/signer.go:NewCryptoSigner()` so if other folks are using that public API they benefit from that check (and return an error if you're trying to do ML-DSA with anything other than SHA512)
 
- then the existing check will fail if we can an error back from `NewCryptoSigner()`

Later, if this gets annoying, we could add a check in `pkg/api/api.go` before we call `NewCryptoSigner()` and log a warning to the user that we're changing their requested settings in order to successfully start.

steiza (commented):
I think we just have a minor issue to resolve, otherwise this looks good to me!

This is my first review of `sigstore/timestamp-authority`, and it definitely took some time to get up-to-speed.

I made a script that created a key and a cert chain, and was able to run the server from those files, make a request, and verify it. The hash algorithm used in embedded in the request, and so it does not need to be specified separately at verification time (and the TSA already supports SHA-512 verification).

If it's helpful, I can supply the script and the commands I used to test, but I suspect that TSA maintainers are more proficient at this than I am : D

pkg/api/api.go
Hayden-IO, line 101: We should also verify that each public key in the chain is an ML-DSA key.

go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.24.1
 	github.com/rs/cors v1.11.1
-	github.com/sigstore/sigstore v1.10.9
+	github.com/sigstore/sigstore v1.10.10-0.20260910123929-49e21e51fc69
```
Hayden-IO, line 26: Can bump to v1.10.10 now

pkg/api/api.go
Hayden-IO, line 1: Can we have an end-to-end test as well, in https://github.com/sigstore/timestamp-authority/blob/main/pkg/tests/api_test.go or a new file?

Hayden-IO (changes requested):
Can you put support for this behind a feature flag? ~We don't want support for this rolling out to the current production instance, we'll have a separate instance for PQC. Otherwise we'll end up with PQC signatures counter-signed by a non-PQC authority.~

~The converse - a PQC authority counter-signing non-PQC signatures - is an option, though I'd start with keeping PQC and non-PQC disjoint with a simple `--enable-pqc` flag. If that's needed in the future, we can add a supported algorithms flag.~

When the feature flag is set, we should enforce on startup that the server signing key uses ML-DSA and all public keys in the chain are for ML-DSA, ~and for each request, that any signature is for ML-DSA.~

Edit: NVM, signature data isn't provided in the TS request. But nonetheless, I think it'd be good to have this behind a feature flag to reserve the right to make breaking changes.

pkg/api/api.go
Hayden-IO: Unfortunate that RFC9882 didn't go with SHAKE256 to match the default for FIPS 204, but such is life. I'd keep the set of supported hash algs to a minimum for now. If SHA-512 is the only MUST, let's stick with that, and +1 to throwing an error if something else is provided.

